Issue #321, opened 2015-01-19
Bug in netcdftime.DateFromJulianDay


jhamman:
Looking at this code in [`DateFromJulianDay()`](https://github.com/Unidata/netcdf4-python/blob/0ec844a95fe1294dbc7bec9c3796cf07eb7af008/netcdftime/netcdftime.py#L311-L313):
```python
...
minute = np.clip((F * 1440. + eps).astype(np.int64), 0, 59)
second = np.clip((F - minute / 1440.) * 86400., 0, None)
second = second.astype(np.int32)
microsecond = ((second - np.int32(second)) * 1e6).astype(np.int32)
...
```
Should the order of the last two lines be switched since casting second to `np.int32` prior to calculating `microsecond` would result in a zero for all cases?  Essentially, the code reads this `microsecond = ((np.int32(second) - np.int32(second)) * 1e6).astype(np.int32)`

jswhit:
Good catch - thanks!  Fix in pull request #322

jswhit:
Turns out the microseconds calculation is not really used for anything (internally at least).  If the units are in microseconds, python datetime is used. It's a good thing too, since I don't think the microseconds computation is correct.  For example

    from netcdftime import DateFromJulianDay, JulianDayFromDate
   from datetime import datetime
    d = datetime(1858, 11, 17, 0, 1, 1, 0)
    print d
    mjd = JulianDayFromDate(d,calendar='proleptic_gregorian')
    print mjd
    date = DateFromJulianDay(mjd)
    print date

gives

    1858-11-17 00:01:01
    2400000.50071
    1858-11-17 00:01:00.999990

after your fix.

jhamman:
Interesting.  I think you are right about it being a precision issue since `mjd` is stored as a float in units of days.

jswhit:
It's a known issue. One solution is to use a 2-tuple of floats to represent the Julian day, as in jdcal (https://pypi.python.org/pypi/jdcal).  Unfortunately, jdcal only supports the proleptic gregorian and Julian calendars.

jswhit:
microseconds is no l

PR #322, opened 2015-01-20
fix for microsecond calculation in netcdftime (issue #321)

jswhit:


jhamman:
Looks good.  Thanks @jswhit .

Issue #325, opened 2015-02-02
indexing very very slow since  v1.1.2


s-hahn:
In the changelog it has been mentioned that

* indexing netCDF variables with integer sequences and boolean arrays now
behave the same way (integer sequences are converted to boolean arrays
internally). Addresses issue #300. 

I've tested indexing the same data with v1.1.1 and v1.1.2 (about 220000 elements) with very scattered int indices (e.g. 2000, 2300, 10002, 100000, .. overall 150 indices) and the performance difference is really shocking.

jswhit:
Would you mind posting a code snippet with timing results?

s-hahn:
This test shows that v1.1.1 is faster about 130 times. 

np.unique is actually only needed for v1.1.2, which requires sorted indicies without duplicates.

netCDF4 version: 1.1.2
write function took 4704.421 ms
read function took 12191.375 ms

netCDF4 version: 1.1.1
write function took 5326.627 ms
read function took 89.266 ms

```python
import time
import netCDF4
import numpy as np

def timing(f):
    """
    Wrapper for function run-time.
    """
    def wrap(*args):
        time1 = time.time()
        ret = f(*args)
        time2 = time.time()
        print('%s function took %0.3f ms' %
              (f.func_name, (time2 - time1) * 1000.0))
        return ret
    return wrap


@timing
def write(fn, var1):
    with netCDF4.Dataset(fn, mode='w') as nc:
        nc.createDimension('dim1')
        nc_var1 = nc.createVariable('var1', var1.dtype,
                                    dimensions=('dim1'))
        nc_var1[:] = var1


@timing
def read(fn, s):
    subset = np.unique((np.random.randint(0, s, 2000)))
    with netCDF4.Dataset(fn, mode='r') as nc:
        x = nc.variables['var1'][subset]
    return x

print("netCDF4 version: {:}".format(netCDF4.__version__))
fn = 'test.nc'
var1 = np.arange(500000)
write(fn, var1)
read(fn, var1.size)
```

jswhit:
The slowness comes from the block of code in netCDF4_utils.py (around line 246) that tries to convert the indices to a slice object.  If that code is bypassed, there is no difference in speed between 1.1.1 and 1.1.2.

jswhit:
Pull request #326 fixes this by not trying to convert boolean indices to slices except for the very simplest case (all elements in the range are True).

jswhit:
@shoyer , I like your solution much better.

PR #326, opened 2015-02-02
eliminate expensive conversion of boolean indices to slice objects (issue #325)

jswhit:


netCDF4_utils.py
```diff
@@ -244,36 +244,37 @@ def _StartCountStride(elem, shape, dimensions=None, grp=None, datashape=None,\
...
-                            if e[i]: step = False
-                        else:
-                            if not e[i]: step = False
-                if step: # it step False, can't convert to slice.
-                    newElem.append(slice(start,stop,step))
-                else:
-                    newElem.append(e)
-            else:
-                newElem.append(slice(0,0))
+           el = e.tolist()
```
shoyer: This is going to be a slow step regardless -- we really don't want to convert to lists if possible to avoid.

shoyer:
@jswhit I'm going to submit an alternate version of this PR with vectorized checks.

PR #335, opened 2015-02-16
fix reading of scalar vlen string variables

jswhit:


jswhit:
Turns out scalar vlens do work, they are just not read correctly.  This pull request now fixes the reading bug instead of disallowing scalar vlens.

Issue #330, opened 2015-02-12
num2date returning incorrect values


dariuspbom:
num2date appears to be returning incorrect values frequently off by one.

Underlying lib is netcdf-4.3.2.  Python netCDF4 versions 1.1.2 and 1.1.3 appear to be affected.  Version 1.1.1 is good.

Demonstrated with the following:

>>> import netcdftime
>>> t = netcdftime.utime("seconds since 1970-01-01T00:00:00Z")
>>> for n in xrange(60):
...     print n, t.num2date(n), int(round(t.date2num(t.num2date(n))))
... 
0 1970-01-01 00:00:00 0
1 1970-01-01 00:00:00 0
2 1970-01-01 00:00:01 1
3 1970-01-01 00:00:02 2
4 1970-01-01 00:00:04 4
5 1970-01-01 00:00:05 5
6 1970-01-01 00:00:06 6
7 1970-01-01 00:00:07 7
8 1970-01-01 00:00:07 7
9 1970-01-01 00:00:08 8
10 1970-01-01 00:00:09 9
11 1970-01-01 00:00:10 10
12 1970-01-01 00:00:12 12
13 1970-01-01 00:00:13 13
14 1970-01-01 00:00:14 14
15 1970-01-01 00:00:14 14
16 1970-01-01 00:00:15 15
17 1970-01-01 00:00:16 16
18 1970-01-01 00:00:17 17
19 1970-01-01 00:00:19 19
20 1970-01-01 00:00:20 20

Only up to 20 shown, but you can see that number 1, 2, 3, 8, 9, 10, 11, 15, 16, 17, 18 all get converted to a date with num2date incorrectly.

With version 1.1.1 the output is good:

>>> import netcdftime
>>> t = netcdftime.utime("seconds since 1970-01-01T00:00:00Z")
>>> for n in xrange(60):
...     print n, t.num2date(n), int(round(t.date2num(t.num2date(n))))
... 
0 1970-01-01 00:00:00 0
1 1970-01-01 00:00:01 1
2 1970-01-01 00:00:02 2
3 1970-01-01 00:00:03 3
4 1970-01-01 00:00:04 4
5 1970-01-01 00:00:05 5
6 1970-01-01 00:00:06 6
7 1970-01-01 00:00:07 7
8 1970-01-01 00:00:08 8
9 1970-01-01 00:00:09 9
10 1970-01-01 00:00:10 10
11 1970-01-01 00:00:11 11
[...]

PR #336, opened 2015-02-16
should fix issue #330 while keeping sub-second accuracy

cpaulik:
This fixes issue #330 on my install. It also changes the _datetime extension and adds microsecond support to it.


test/tst_netcdftime.py
```diff
@@ -117,7 +117,7 @@ def runTest(self):
         d2 = self.cdftime_noleap.num2date(t1)
         self.assertTrue(str(d1) == str(d2))
         # check day of year.
-        ndayr = d2.timetuple()[7]
+        ndayr = d2.timetuple()[6]
```
cpaulik: something is wrong here, this should not work

jswhit:
used pull request #334 instead.  If we want microsecond accuracy, we'll have to revisit the algorithm. Perhaps something like jdcal (https://github.com/phn/jdcal), which stores the julian day as two 64 bit floats.

Issue #340, opened 2015-02-20
Install failed - netCDF4 v1.1.4


duytrung:
I installed the package and was not successful. Could you please, provide the package in Python Wheel standard. Thanks!
Here is the log:
```
<code>

c:\PYX>pip install -U netCDF4
Collecting netCDF4 from https://pypi.python.org/packages/source/n/netCDF4/netCDF4-1.1.4.tar.gz#md5=0
c9d6c2f625e20855282bc0d5bf12a8c
  Using cached netCDF4-1.1.4.tar.gz
    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "c:\users\myexpe~1\appdata\local\temp\pip-build-qomxw5\netCDF4\setup.py", line 187, in <m
odule>
        retcode =  subprocess.call([ncconfig,'--libs'],stdout=subprocess.PIPE)
      File "C:\Pyx\lib\subprocess.py", line 522, in call
        return Popen(*popenargs, **kwargs).wait()
      File "C:\Pyx\lib\subprocess.py", line 710, in __init__
        errread, errwrite)
      File "C:\Pyx\lib\subprocess.py", line 958, in _execute_child
        startupinfo)
    WindowsError: [Error 2] The system cannot find the file specified
    Complete output from command python setup.py egg_info:
    reading from setup.cfg...

    Traceback (most recent call last):

      File "<string>", line 20, in <module>

      File "c:\users\myexpe~1\appdata\local\temp\pip-build-qomxw5\netCDF4\setup.py", line 187, in <m
odule>

        retcode =  subprocess.call([ncconfig,'--libs'],stdout=subprocess.PIPE)

      File "C:\Pyx\lib\subprocess.py", line 522, in call

        return Popen(*popenargs, **kwargs).wait()

      File "C:\Pyx\lib\subprocess.py", line 710, in __init__

        errread, errwrite)

      File "C:\Pyx\lib\subprocess.py", line 958, in _execute_child

        startupinfo)

    WindowsError: [Error 2] The system cannot find the file specified

    ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in c:\users\myexpe~1\appdata\local\t
emp\pip-build-qomxw5\netCDF4

</code>
```
Regards

matteobachetti:
[...]

PR #342, opened 2015-02-20
Fix issue in installation where USE_NCCONFIG is requested but the nc-config is not installed

swails:
This addresses #340.

This just started on some Travis systems because version 1.1.4 introduced a `use_ncconfig = True` in the default `setup.cfg` file that pip was executing.  This caused `setup.py` to look for `nc-config` and try to run it, hoping it was in the `PATH`.  If it *wasn't*, it threw an `OSError` rather than simply returning a non-zero exit code.

On Windows, this raised a `WindowsError`, whereas on *nix it raised a `OSError`.  Since I'm not sure if `WindowsError` subclasses `OSError` and I don't use Windows, I just threw in a bare except (I don't think there's a real risk of that masking an error in this case).

A workaround is to make sure that `nc-config` *is* available when you build `netCDF4-1.1.4` (on Travis, that means installing `netcdf-bin` in addition to `libnetcdf-dev`).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -184,7 +184,10 @@ if USE_NCCONFIG is not None:
             ncconfig = os.path.join(netCDF4_dir,'bin/nc-config')
         else: # otherwise, just hope it's in the users PATH.
             ncconfig = 'nc-config'
-    retcode =  subprocess.call([ncconfig,'--libs'],stdout=subprocess.PIPE)
+    try:
+        retcode =  subprocess.call([ncconfig,'--libs'],stdout=subprocess.PIPE)
+    except:
+        retcode = 1
 else:
     retcode = 1
 


jswhit:
The other option would be to remove the setup.cfg file from the tarball.  I didn't really  intend for it to be included anyway.  I'll go ahead and do that while we are considering this PR.

swails:
Version 1.1.3 also had a `setup.cfg` file, so I thought that was common practice.

Although I think the aim of the chunk of code here is to see if `nc-config` can actually be used to help in the installation of netCDF4.  It appears (to me, at least), that the intent of this part of setup.py is to see if nc-config can actually be used to help.  If it doesn't exist, I figured the better response was "no, it can't help", rather than "where is it?? let's crash the install".

dopplershift:
I agree--I think using `nc-config` by default is great, but it should fail gracefully, not error out.

jswhit:
version 1.1.3 did have a setup.cfg file, but it did not have nc_config=True in it.

jswhit:
merging now, will set use_ncconfig=True in next release

swails:
Great!  Thanks for the quick fix!

PR #356, opened 2015-03-03
fix for issue 353

jswhit:


test/tst_netcdftime.py
```diff
@@ -351,6 +351,9 @@ def runTest(self):
                 err = numpy.abs(days1 - days2)
                 assert(err < eps)
                 assert(date1.strftime(dateformat) == date2.strftime(dateformat))
+    # issue 353
+    assert (num2date(0, 'hours since 2000-01-01 0') == \
```
shoyer: style note: you don't need `\` to mark continued lines when you have an open parenthesis

shoyer:
Thanks!

PR #370, opened 2015-03-12
release GIL for reading of netcdf files (issue 369)

jswhit:


jswhit:
This call wraps all netcdf C lib calls that pertaining to reading "with nogil:" in cython.  This releases the GIL.

Issue #385, opened 2015-03-24
Runtime detection of indexing semantics


rhattersley:
netCDF4 and numpy differ in how they treat tuple/array indexing arguments:
```python
>>> type(v)
<type 'netCDF4.Variable'>
>>> v.shape
(240, 37, 49)
>>> v[(0, 1), (0, 2, 3)].shape
(2, 3, 49)
>>> np.array(v)[(0, 1), (0, 2, 3)].shape
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
IndexError: shape mismatch: indexing arrays could not be broadcast together with shapes (2,) (3,) 
```

It's something of a long shot, but ... given an object that supports indexing, how can I determine at runtime which of these two styles it supports?

I like to avoid `isinstance(v, netCDF4.Variable)`, or probing the behaviour with something like `v[(0, 1), (0, 2, 3), 0]` as it could have a relatively large cost.

jswhit:
hasattr(v,'_varid') ?

shoyer:
The best I've come up with is using isinstance checks. Which actually seems to work OK, for the most part. I've found most stores support neither type of fancy indexing with multiple arrays, unless they are just proxying indexing on to a library like netCDF4 or numpy.

The other option is coming is up with some sort of convention for how to mark "orthogonally" indexed arrays such as some special attribute.

rhattersley:
> The best I've come up with is using isinstance checks.

That's what I'm inclined to use, unless ...

> The other option is coming is up with some sort of convention for how to mark "orthogonally" indexed arrays

... this gets some traction. But that would need to involve other libraries. Does xray support one (or both?) of these indexing styles?

dopplershift:
Couldn't we just standardize on what numpy itself supports?

```python
import numpy as np
a = np.zeros((240,37,49))
print a[np.array([0,1])[:,None], np.array([0,2,3])].shape
```
prints `(2,3,49)` for me. Essentially, instead of tuples, you need to pass in arrays that can broadcast against each other.

jswhit:
Or just

    print a[(0,1),:,:][:,[0,2,3]].shape

This works for netcdf variables too, but it is probably a lot slower since it reads more data.

    import numpy as np
    from netCDF4 import Dataset
    a = np.zeros((240,37,49))
    print a[(0,1),:,:][:,[0,2,3]].shape
    nc = Dataset('test.nc','w')
    xd = nc.createDimension('x',240)
[...]

PR #386, opened 2015-03-25
allow integer index arrays to be multi-dimensional as long

jswhit:
as they can be 'squeezed' to 1d.  Partially addresses issue #385.


shoyer:
I am -1 on this change because it allows for indexing like numpy that does not obey numpy indexing rules, e.g., see
```
In [13]: x = np.arange(4).reshape(2, 2)

In [14]: x[[[0], [1]], [[0, 1]]]
Out[14]:
array([[0, 1],
       [2, 3]])

In [15]: x[[[0, 1]], [[0], [1]]]
Out[15]:
array([[0, 2],
       [1, 3]])
```

jswhit:
Fixed this, so slicing a netcdf variable as like this

x[[[0, 1]], [[0], [1]]]

will fail (since otherwise it would give a different answer than numpy).

However, it still differs in the sense that

x[[[0]],[[1]]] 

will return a (1,1) array in numpy and a float in netcdf4-python.

jswhit:
Closing this, since @shoyer convinced me this is not a good idea.

PR #387, opened 2015-03-31
add __orthogonal_indexing__ Variable attribute (issue #385).

jswhit:


netCDF4.pyx
```diff
@@ -2465,6 +2465,8 @@ instance. If C{None}, the data is not truncated. """
         cdef size_t sizep, nelemsp
         cdef size_t *chunksizesp
         cdef float preemptionp
+        # flag to indicate that orthogonal indexing is supported
+        self.__orthogonal_indexing__ = True
```
shoyer: This could be a class attribute instead of an instance attribute, e.g., you could just set `__orthogonal_indexing__ = True` at one higher level.

jswhit: Added Dataset.__orthogonal_indexing__ and Group.__orthogonal_indexing__

rhattersley:
Thanks @jswhit :+1:

PR #389, opened 2015-04-07
write unicode attributes as strings (using nc_put_att_string).

jswhit:


shoyer:
Looks great! Have you tested that it actually works?

shoyer:
It looks like this is already tested implicitly on Python 3, which is why the first tests with Travis failed.

PR #392, opened 2015-04-09
covert scripts to entry_points

mindw:
windows friendly scripts.


jswhit:
I'm not familiar with 'entry points' in setuptools.  Can you explain why this makes the scripts 'windows friendly'?

mindw:
... On Windows, a set of foo.exe, bar.exe, and baz.exe launchers are created, alongside a set of foo.py, bar.py, and baz.pyw files. The .exe wrappers find and execute the right version of Python to run the .py or .pyw file ...

https://packaging.python.org/en/latest/distributing.html#console-scripts
https://pythonhosted.org/setuptools/setuptools.html#automatic-script-creation

jswhit:
Seems reasonable - but it does require setuptools, which is not currently required to build netcdf4-python.

mindw:
https://packaging.python.org/en/latest/current.html#id6


Even for projects that do choose to use distutils, when *pip*
<https://packaging.python.org/en/latest/projects.html#pip> installs such
projects directly from source (rather than installing from a prebuilt
*wheel* <https://packaging.python.org/en/latest/glossary.html#term-wheel>
file),
it will actually build your project using *setuptools*
<https://packaging.python.org/en/latest/projects.html#setuptools> instead.


-gabi

mindw:
In any case - the latest version is distutils friendly

jswhit:
OK, great.  Merging now - thanks for the contribution.

PR #396, opened 2015-04-11
gitattributes to limit export

zerothi:
Adding attributes to limit the export of several _huge_ files.

Also the hidden files are strictly not needed.


zerothi:
Not needed anymore...

PR #399, opened 2015-04-15
Remove code that has no effect

mdunphy:
Remove this line because the next three lines replace its action.
(Actually in my case that line causes python to error out, and I can't see any justification for keeping it.)


jswhit:
Thanks - merging now...

Issue #404, opened 2015-04-29
Update minimum required numpy version and add it to Travis-CI tests


shoyer:
As reported by @matthew-brett in #294, tests for netCDF4-python do not seem to pass on older versions of NumPy -- despite setup.py reporting a minimum required version of 1.3.

We should figure out what the oldest version of numpy is that can run this library successfully, and then add a test for that configuration to Travis-CI to ensure that we don't have unexpected regressions.

matthew-brett:
For testing minimal dependencies - this is what I do for one of my projects: https://github.com/nipy/nibabel/blob/master/.travis.yml

The basic idea is having an environment variable in `.travis.yml` called 'DEPENDS' which usually has:

```
DEPENDS="numpy scipy matplotlib h5py"
```

but for one element in the travis-ci test matrix has:

```
DEPENDS="numpy==1.5.1 pydicom==0.9.7"
```

which are our declared minimum dependencies.

jswhit:
Thanks for that suggestion.  I've fixed things to tests all pass with numpy 1.7.0, and added testing of minimal dependecies (both numpy and cython) to .travis.yml in pull request #405.

jswhit:
pull request #405 merged, making numpy 1.7.0 the minimum required.

shoyer:
awesome, thanks!

PR #405, opened 2015-05-02
test for minimium dependencies (issue #404)

jswhit:


shoyer:
It looks like setup.py could also use an updated numpy version?

jswhit:
right - setup.py now updated.

PR #414, opened 2015-05-12
createVariable and createGroup take unix-like paths

jswhit:
motivated by a new feature in h5netcdf, createVariable and createGroup now take paths as arguments, so you can do this:

    ﻿from netCDF4 import Dataset
    f = Dataset('test.nc','w')
    x = f.createDimension('x',10)
    # create groups in path if they don't already exist
    v = f.createVariable('/grouped/data/v',float,('x',))
    g = f.groups['grouped']
    # create groups underneath 'grouped'
    v2 = g.createVariable('/data/data2/v2',float,('x',))
    # the next two lines do the same thing as above
    #v2 = g.createVariable('./data/data2/v2',float,('x',))
    #v2 = g.createVariable('data/data2/v2',float,('x',))
    print f['/grouped/data'] # returns group instance
    print f['/grouped/data/data2/v2'] # returns variable instance
    f.close()

Since forward slashes were not allowed in variable and group names anyway, this should not introduce any ambiguity.  However, the behavior of createGroup changes so that``nc. createGroup(groupname)`` will no longer fail if ``groupname`` already exists.  Instead, the existing group is returned.  As a result, code that relies on ``createGroup`` failing if the group already exists will break.


shoyer:
This looks great! Glad to see this making it back to netCDF4-python :).

PR #418, opened 2015-05-20
convert docstrings from epytext markup to markdown

jswhit:
so pdoc (https://github.com/BurntSushi/pdoc) can be used to generate docs instead of epydoc (which is no longer maintained).  Will make is easier to migrate to Sphinx or readthedocs in the future, since both can interpret markdown.  

See http://htmlpreview.github.io/?https://github.com/Unidata/netcdf4-python/blob/pdoc/docs/netCDF4/index.html for what the pdoc docs look like, compared to epydoc (http://unidata.github.io/netcdf4-python/)


shoyer:
Can Sphinx really interpret Markdown? In my experience, it only does reStructuredText.

(Note that ReadTheDocs uses Sphinx.)

jswhit:
Apparently it is possible to use markdown with sphinx:

http://stackoverflow.com/questions/2471804/using-sphinx-with-markdown-instead-of-rst

and also with readthedocs:

http://docs.readthedocs.org/en/latest/getting_started.html#in-markdown

shoyer:
Looks like Markdown may be a viable option, though those stackoverflow responses on Sphinx don't look very encouraging to me.

In any case, the new docs do look much nicer than the old ones! The only thing that would be nice to see is a table of contents for the tutorial.

jswhit:
Unfortunately, neither sphinx or pdoc extracts the docstrings for properties.  I used the __pdoc__ mechanism in pdoc to work around this.

shoyer:
Hmm. In the long term, that might be a good reason to follow the lead of h5py, and write the higher level API in Python around a low level Cython wrapper for the netCDF-C library.

jswhit:
I just can't get excited the idea about adding a extra layer of python just to allow the docstrings to be more easily extracted.

shoyer:
Well, there would be a few other benefits, too: it would also make it easier to debug when things go wrong and there could be far less generated C code (right now it's with every PR). Still, it's a lot of work with a questionable return.

jswhit:
Added a table of contents for the tutorial.

PR #427, opened 2015-06-04
Don't compile netCDF4.pyx on clean.

sebastic:
The clean target shouldn't modify files include in the source tree.


jswhit:
make sense, thanks.

PR #428, opened 2015-06-04
Don't run tests that require network connectivity if NO_NET is set.

sebastic:
This excludes tst_dap.py from being run to allow running all remaining tests in build chroots without network.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 env:
     global:
         - DEPENDS="numpy cython"
+        - NO_NET=1
 python:
   - "2.7"
   - "3.3"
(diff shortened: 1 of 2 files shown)


sebastic:
The Travis CI build now fails because it also cannot run tst_dap.py, if you apply this PR and set the NO_NET environment variable the tests will succeed.

jswhit:
The Travis CI test sometimes fails, not because there is no network, but because the server the test uses is down.  I suppose it's useful to be able to suppress this test with an env var, so I'll go ahead and merge.  Thanks for the flurry of pull requests!

sebastic:
> Thanks for the flurry of pull requests!

Thanks for applying them so quickly.

They were all triggered by the update of the Debian package to 1.1.8 (still in the [NEW queue](https://ftp-master.debian.org/new/netcdf4-python_1.1.8-1.html) unfortunately).

PR #435, opened 2015-07-15
fixed auto_scale for unsigned int

hvwaldow:
We are packing floats into into unsigned integers instead of signed integers.  Unsigned integers are supported in NetCDF-4. Netcdf4-python doesn't round to convert the scaled floats into integers (for Variable.scale == True)  if the target type is an unsigned int. This results in errors that are biased negative and up to 2 times larger than the ones obtained by correctly rounding.


jswhit:
Looks reasonable, thanks.  Could you add a Changelog entry before I merge?  Adding a test to test/tst_scaled.py would be nice too.

hvwaldow:
Haven't figured out why this fails on Travis but tests OK locally. Will get back to it tomorrow.

hvwaldow:
It appears that cython needs to be installed so that writing / reading unsigned int works. I guess that netCDF 4.1.1 has a problem with unsigned integers. Might be related to:

http://www.unidata.ucar.edu/software/netcdf/docs/release_notes.html
<b>4.1.3 2011-06-17</b>

+ Define uint type for unsigned integer, if not otherwise available. Symptom was compile error involving uint in putget.c. [NCF-49]

I don't know how to deal with that. Do you have a suggestion?

jswhit:
I don't see how having cython installed would affect that test - it's just going to trigger regeneration of _netCDF.c from _netCDF4.pyx.  I see the tests are passing now, maybe it was a transient issue with travis?

jswhit:
Oh, I see you removed the unsigned integer version of the tests.

jswhit:
I think the problem is that you did not regenerate _netCDF.c from the updated _netCDF.pyx.  That means that when you run without cython, your updates are not part of the resulting _netCDF4.so.  To regenerate _netCDF4.c, cd into the netCDF4 directory and run 'cython -I ../include _netCDF4.pyx' after first making sure that include/constants.pyx has all variables set to zero.

hvwaldow:
Ok. That was it! Thanks!

Note on the new testcase: I noted that auto_scaled writing was not tested at all before. So the new test is somewhat orthogonal to the rest of tst_scaled.py. It compares original_values with auto_unscaled(read(write(auto_scaled(original values)))) but ignores interaction with auto_mask.

jswhit:
merging now...

Issue #444, opened 2015-07-22
integer overflow in date2index()


richli:
Hi,

Here's a minimal example that triggers a RuntimeWarning about integer overflow:

```python
from datetime import datetime

import numpy as np
from netCDF4 import Dataset
from netCDF4 import num2date, date2index

with Dataset("test.nc", "w") as f:
    f.createDimension("record", 5)
    f.createVariable("time", np.int32, ("record", ))
    f['/time'].units = "seconds since 1970-01-01 00:00:00"

    f['/time'][:] = [1419602300, 1419602400, 1419603000, 1419603600, 1419604200]
    print(num2date(f['/time'][:], f['/time'].units))

    query_time = datetime(2014, 12, 26, 14, 15, 39, 531250)
    print(query_time)
    index = date2index(query_time, f['/time'], select='nearest')
    print(index)
```

Here's the output:

```
$ python ./test.py                                                                                      
[datetime.datetime(2014, 12, 26, 13, 58, 20)
 datetime.datetime(2014, 12, 26, 14, 0)
 datetime.datetime(2014, 12, 26, 14, 10)
 datetime.datetime(2014, 12, 26, 14, 20)
 datetime.datetime(2014, 12, 26, 14, 30)]
2014-12-26 14:15:39.531250
/usr/lib/python3.4/site-packages/netcdftime/netcdftime.py:1087: RuntimeWarning: overflow encountered in int_scalars
  [nctime[i - 1] + nctime[i] for i in index[mismatch]]) / 2.
3
```

(In this case it selects the correct index of 3, but in my full example it instead returns 0).

I examined where the overflow occured (https://github.com/Unidata/netcdf4-python/blob/v1.1.8rel/netcdftime/netcdftime.py#L1086). It looks like the code is taking the sample mean of two large integers, but upon addition of the two integers, it overflows. As a workaround, if I cast from int32 to a int64, this no longer overflows. Another workaround is to shift my epoch by a few decades to decrease the time values (e.g., shifting from 1970-01-01 to 2000-01-01).

Thanks! I'm using Python 3.4.3, netcdf4-python 1.1.8, netcdf 4.3.3.1, and hdf5 1.8.14.

jswhit:
You really shouldn't be using 32 bit integers to represent ``"seconds since 1970-01-01 00:00:00"``.  Changing the type of your time variable to np.int64 should solve the problem.

richli:
[...]

PR #445, opened 2015-07-24
fix for integer overflow in date2index (issue #444)

jswhit:


shoyer:
could we please add a regression test?

jswhit:
Done.

PR #454, opened 2015-09-10
Add a convenience method for quick variable filtering

ocefpaf:
@jswhit Is there an interest for a `get_variables_by_attributes` method?  This version is stolen from [pyaxiom](https://github.com/axiom-data-science/pyaxiom/blob/master/pyaxiom/netcdf/dataset.py#L10-L47)'s `EnhancedDataset`.

For example:
```python
u = nc.get_variables_by_attributes(standard_name='eastward_sea_water_velocity')
v = nc.get_variables_by_attributes(standard_name='northward_sea_water_velocity')
```

Pinging @kwilcox, who is the original author of the method, and  @rsignell-usgs who is also interested into having this functionality.

If this is useful here I will write do not merge.  I will better `docstrings` for it.


lesserwhirls:
This is an extremely useful method in netCDF-Java! +1

rsignell-usgs:
As @ocefpaf mentioned, the CF community folks (and many others) would find this very useful!

dopplershift:
Seems reasonable to me.

jswhit:
Looks great, thanks!  Needs a Changelog entry and a test though.

ocefpaf:
@jswhit I have two quick questions:

1) What are the instructions for the non-cython test?  Should I re-generate the `_netCDF4.c
` with `python setup.py build` and commit it?

2) The tests are not very comprehensive.  They just run the `get_variables_by_attributes` method, using a combination of possibilities, in a known dataset that is available online. Is that OK?  (I guess that the right way should be mocking a `nc` object, but I do not know where to start.)

test/tst_get_variables_by_attributes.py
```diff
@@ -0,0 +1,46 @@
+import unittest
+
+import netCDF4
+
+class VariablesByAttributesTests(unittest.TestCase):
+
+    def setUp(self):
+        netcdf_file = ('http://geoport.whoi.edu/thredds/dodsC/'
```
shoyer: I would strongly suggest creating a netCDF file to test instead. In my experience, remote DAP servers are not always reliable. This is especially annoying for packagers which run your unit tests to verify that they built your project successfully.

jswhit:
To regenerate _netCDF4.c, edit include/constants.pyx and make sure all values are set to zero (or just remove the file and check it out again from git).  Then, in the _netCDF directory, run ``cython -I ../include _netCDF4.pyx``.

I agree with @shoyer about not using opendap for a test.  You could just create your own simple file, and the exercise ``get_variable_by_attributes`` by reading it.  It doesn't have to be comprehensive.

ocefpaf:
`_netCDF4.c` regenerated.  Working on the test file now...

ocefpaf:
... All done.  Let me know if this is OK.

jswhit:
Looks great - thanks a lot!  Merging now...

Issue #453, opened 2015-09-10
date2index broken


FrancescAlted:
I am experiencing a strange error when using date2index in latest netCDF4 1.1.9 (via Anaconda on Ubuntu 15.04).  Here it is:

```
In [44]: nc = netCDF4.Dataset("/home/faltet/uio/shyft-data/netcdf/arome-testdata/arome_metcoop_red_default2_5km_20150823_06.nc")

In [45]: tvar = nc.variables['time']

In [46]: tvar[:3]
Out[46]: array([  1.44030960e+09,   1.44031320e+09,   1.44031680e+09])

In [47]: dt = datetime.datetime.utcfromtimestamp(int(1.44031320e+09))

In [48]: netCDF4.date2index(dt, tvar)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-48-e2a14096dc8f> in <module>()
----> 1 netCDF4.date2index(dt, tvar)

netCDF4/_netCDF4.pyx in netCDF4._netCDF4.date2index (netCDF4/_netCDF4.c:51152)()

netCDF4/_netCDF4.pyx in netCDF4._netCDF4.Variable.__getattr__ (netCDF4/_netCDF4.c:30655)()

netCDF4/_netCDF4.pyx in netCDF4._netCDF4.Variable.getncattr (netCDF4/_netCDF4.c:28214)()

netCDF4/_netCDF4.pyx in netCDF4._netCDF4._get_att (netCDF4/_netCDF4.c:3548)()

AttributeError: NetCDF: Attribute not found

In [49]: netCDF4.__version__
Out[49]: '1.1.9'
```

I'd expect getting just a 1.

jswhit:
date2index is trying to access the calendar attribute of the time variable, since the calendar kwarg was not given.   If the calendar is one that python datetime supports, then python datetime is used, otherwise the netcdftime module is used.  It should either fail with a more informative error message, or just use netcdftime if it can't decide whether python datetime will work.  The only advantage of using python datetime is greater accuracy (millseconds or less, netcdftime has only 1 second accuracy).

jswhit:
Actually, it's the 'units' attribute that it is choking on, not 'calendar'.

FrancescAlted:
Well, the user's guide is saying this (http://netcdf4-python.googlecode.com/svn/trunk/docs/netCDF4-module.html#date2index):

```
[...]

PR #456, opened 2015-09-12
check for units attribute in date2index (issue #453)

jswhit:


FrancescAlted:
LGTM.  Also, you may want to replace this docstring:

"""
If calendar is None, its value is given by nctime.calendar or standard if no such attribute exists.
"""

by

"""
If calendar is None, its value is given by nctime.calendar, raising an error if no such attribute exists.
"""

or similar.

jswhit:
Why?  It does assume 'standard' if there is no ``calendar`` attribute or kwarg.

netcdftime/netcdftime.py
```diff
@@ -1002,6 +1002,10 @@ def date2index(dates, nctime, calendar=None, select='exact'):
     an exact match cannot be found. C{nearest} will return the indices that
     correpond to the closest dates.
     """
+    try: 
+        nctime.units
+    except AttributeError:
+        raise AttributeError("netcdf time variable is missig a 'units' attribute")
```
FrancescAlted: missig -> missing

netcdftime/netcdftime.py
```diff
@@ -1036,6 +1040,10 @@ def time2index(times, nctime, calendar=None, select='exact'):
     an exact match cannot be found. C{nearest} will return the indices that
     correpond to the closest times.
     """
+    try: 
+        nctime.units
+    except AttributeError:
+        raise AttributeError("netcdf time variable is missig a 'units' attribute")
```
FrancescAlted: missig -> missing

FrancescAlted:
You are right:  I was reading the docs badly.

Issue #452, opened 2015-09-10
Variables with a enum data type are unsupported


joaquinrgu:
Variables with a enum data type are not recognized.

The test file cloud2.nc (https://drive.google.com/folderview?id=0B6RTzwJWtu6zOFpDQjFSd2U0clU&usp=sharing) was created with ncgen using the following CDL file (based on http://www.unidata.ucar.edu/software/netcdf/workshops/2011/groups-types/EnumCDL.html) : 

netcdf cloud2 {
types:
  byte enum cloud_t {Clear = 0, Cumulonimbus = 1, Stratus = 2, 
      Stratocumulus = 3, Cumulus = 4, Altostratus = 5, Nimbostratus = 6, 
      Altocumulus = 7, Missing = 127} ;
dimensions:
	time = UNLIMITED ; // (5 currently)
variables:
	cloud_t primary_cloud(time) ;
		cloud_t primary_cloud:_FillValue = Missing ;
        float temperature(time) ;
data:

 primary_cloud = Clear, Stratus, Cumulus, _, Cumulonimbus ;
 temperature = 11,12,13,14,15;
}

It contains two variables: one float and one enum. The float variable works fine, but the enum var returns the following error:
KeyError: 'primary_cloud'

import netCDF4
rootgrp = netCDF4.Dataset('C:\\Users\\rodriguezgu\\NetCDF4\\cloud2.nc', 'r', format='NETCDF4')
print (rootgrp.variables["temperature"])
print (rootgrp.variables["primary_cloud"])
rootgrp.close()

jswhit:
we could support enums the same way h5py does - by mapping to an integer numpy array restricted to certain named values.  See http://docs.h5py.org/en/latest/special.html#enumerated-types.

Would that be useful?

jswhit:
started working on this in the enum branch

joaquinrgu:
Thanks Jeff, that would be useful, in fact the ToolsUI and Panoply viewers are only showing the integer value as far as I know.

jswhit:
enum support appears to be working (pull request #457).  Still need to add docs and a test. Please try it out with your data and report back here.

jswhit:
test/tst_enum.py added.

joaquinrgu:
I confirm I can read now the enum variables in my datasets :)
Thanks also for the test, it shows how to create enum variables.

Maybe now the intro section in http://unidata.github.io/netcdf4-python/ can be updated where it says enums are not supported :)

[...]

PR #457, opened 2015-09-16
add support for enum types (issue #452(

jswhit:


netCDF4/_netCDF4.pyx
```diff
@@ -1186,8 +1257,17 @@ cdef _get_types(group):
...
-                pass
-    return cmptypes, vltypes
+            elif classp == NC_ENUM: # an enum type
+                name = namstring.decode(default_encoding,unicode_error)
+                # read the Enum type info from the file,
+                # create a EnumType instance from it.
+                try:
+                    enumtype = _read_enum(group, xtype)
+                except KeyError:
+                    #print 'WARNING: unsupported Enum type, skipping...'
```
shoyer: When does this happen? Generally it's better to raise exception than pass them silently....

netCDF4/_netCDF4.pyx
```diff
@@ -812,6 +814,60 @@ It is also possible to set contents of vlen string variables with numpy arrays
...
+    <type 'netCDF4._netCDF4.Variable'>
+    enum primary_cloud(time)
+        _FillValue: 127
+    enum data type: uint8
+    unlimited dimensions: time
+    current shape = (5,)
+    >>> print cloud_var[:]
+    [0 2 4 -- 1]
+    >>> nc.close()
+
```
shoyer: Could you add an example of reading the enum dict back from a file on disk?

netCDF4/_netCDF4.pyx
```diff
@@ -1421,6 +1508,10 @@ compound types defined for the `netCDF4.Group` or `netCDF4.Dataset` to instances
 variable-length types defined for the `netCDF4.Group` or `netCDF4.Dataset` to instances 
 of the `netCDF4.VLType` class.
 
+**`enumtypes`**: The `enumtypes` dictionary maps the names of
```
shoyer: My preference would be to *avoid* adding another attribute on the Dataset object. I don't think it's necessary or even helpful, given that all that information is in the dtype.

shoyer:
A few quibbles, but generally looks good to me.

netCDF4/_netCDF4.pyx
jswhit: I used to raise an error for Compound and VLEN types when the read failed, but some users suggested that it would be better to just proceed and read the rest of the file silently ignoring the variables and data types that could not be handled.  Since the potential complexity of HDF5 data types is almost limitless (compound types including vlens of enums..) it is almost certain that there will be files we can't read.

netCDF4/_netCDF4.pyx
jswhit: The 'cmptypes', 'vltypes' and 'enumtypes' attributes are dicts similar to 'dimensions' and 'variables'.  They allow the user to quickly find out what's in the file.  If we didn't provide this, you would have to look at the data type of each variable to tell what kind of exotic data types were being used.

netCDF4/_netCDF4.pyx
jswhit: There is an example in test/tst_enum.py, I could include this code in the docstring.

netCDF4/_netCDF4.pyx
shoyer: I guess if you already have these, it makes sense to add `enumtypes` for consistency. That said -- how else would you find out what sort of exotic types are being used other than looking at the types of each variable? That seems like the most straightforward way to do these things.

netCDF4/_netCDF4.pyx
jswhit: Example added in tutorial.

PR #462, opened 2015-09-24
Fix 'homogenous' typo, replace with 'homogeneous'.

sebastic:
This issue was reported by the lintian QA tool during the Debian package build process.


jswhit:
Thanks!

PR #465, opened 2015-09-24
BUG: String variables can be set using masked arrays

### LINKED ISSUE #464: String variables cannot be set with masked arrays
Trying to set the items of a Variable using a masked array raises a TypeError.  For example:

``` Python
import netCDF4
import numpy as np
dset = netCDF4.Dataset('example.nc', 'w', format='NETCDF4')
dset.createDimension('str_length', 2)
var_u = dset.createVariable('var_u', str, ('str_length', ))
test_u = np.ma.zeros((2, ), dtype='U1')
test_u[:] = np.ma.masked
test_u[0] = 'A'
var_u[:] = test_u[:]
```

```
$ python bug.py 
  File "nc_bug.py", line 11, in <module>
    var_u[:] = test_u[:]
  File "netCDF4/_netCDF4.pyx", line 3905, in netCDF4._netCDF4.Variable.__setitem__ (netCDF4/_netCDF4.c:40014)
    self._put(dataput,a,b,c)
  File "netCDF4/_netCDF4.pyx", line 4130, in netCDF4._netCDF4.Variable._put (netCDF4/_netCDF4.c:41918)
    strdata[i] = data[i]
TypeError: expected string or Unicode object, MaskedConstant found
```

PR to address this will be submitted shortly.

jjhelmus:
closes #464


jjhelmus:
Here the fill value for masked S and U arrays is taken from the default_fillvals, no checking for a missing_value or _FillValue attribute is done which may be more appropriate.  These seems reasonable as no  mask is not applied on non-primitive types in the __getitem__ method.

jswhit:
It's a start, but as you point out it doesn't take into account missing_value or non-default _FillValue.  __getitem__ should also return masked array if we're enabling support for auto masking and unmasking for the str dtype.  This may not be trivial though, since str arrays are returned to python as numpy object arrays, which are odd beasts.

Perhaps it's better to make it more clear that auto-masking is only supported for primitive types.

netCDF4/_netCDF4.pyx
```diff
@@ -3787,6 +3787,8 @@ rename a `netCDF4.Variable` attribute named `oldname` to `newname`."""
                     self._assign_vlen(elem, data)
                     return
                 elif data.dtype.kind in ['S', 'U']:
+                    if ma.isMA(data):
+                        data = data.filled(default_fillvals[data.dtype.kind])
```
jswhit: If the objective is just to avoid a crash here when a masked array is given (and not to provide full auto masking/unmasking support), why not just do

    if ma.isMA(data); data = data.filled()

jjhelmus:
Agreed, I think trying to support assigned of masked arrays to netCDF vlen variables is complicated and is a uncommon edge case.  A error message indicating auto-masking is not supported for non-primitive types makes more sense.

jswhit:
Would you mind modifying your pull request to raise an error message instead of filling the masked array?

jjhelmus:
Done.  A TypeError is now raised when a masked array is used to set a VLEN variable:

``` 
>>> import numpy as np
>>> import netCDF4
>>> dset = netCDF4.Dataset('example.nc', 'w', format='NETCDF4')
>>> dset.createDimension('str_length', 2)
<type 'netCDF4._netCDF4.Dimension'>: name = 'str_length', size = 2

>>> str_var = dset.createVariable('str_var', str, ('str_length', ))
>>> test_ma = np.ma.zeros((2, ), dtype='S1')
>>> str_var[:] = test_ma[:]
  File "<stdin>", line 1, in <module>
  File "netCDF4/_netCDF4.pyx", line 3793, in netCDF4._netCDF4.Variable.__setitem__ (netCDF4/_netCDF4.c:37813)
    raise TypeError(msg)
TypeError: masked arrays cannot be used to assign VLEN str slices.
```

jswhit:
I've merged this pull request in #469

jswhit:
pull request #469 merged, closing...

PR #466, opened 2015-09-25
BUG: fix test teardown on windows.

cournape:
This fixes the test suite on windows: one file was not close, causing the teardown to fail (you cannot remove file which are opened on windows).


jswhit:
Good catch - thanks.

Issue #467, opened 2015-09-29
Allow indexing with unsorted integer arrays


jhamman:
https://github.com/xray/xray/issues/593 has brought indexing with integer arrays back.  There were a few long discussions about this back in #300 and #308 resulting in the integer arrays being required to be sorted and unique. Ultimately, `netCDF4` converts integer arrays to boolean areas for indexing.

In the xray issue above, I have found a reason to support indexing with integer arrays and would like to relax the requirements discussed above.

jswhit:
Can you explain the use case for unsorted integer indices?  Seems like netcdf coordinate variables are almost always going to be monotonic.

jhamman:
@jswhit -

My use case is shown in xray/xray#593 where I was basically sorting a `xray.DataArray` before loading the data from disk.  Since the xray `isel` method returns a view into the data, it was not triggering the loading of the data and eventually, I ended up trying to index into the netCDF4 variable with a unsorted integer array.

This all happened because someone gave me a MERRA Reanalysis dataset that had been poorly concatenated and there were a few scrambled time periods (e.g. [1979, 1980, 1982, 1981]).  Rather than rearranging the time dimension, I tried to sort the dataset using xray.

jswhit:
So the ultimate cause was a malformed dataset.   Is it worth re-introducing un-sorted integer indexing just to work around those cases?

shoyer:
Integer indexing is more general than boolean indexing, so I would argue that it makes more sense to standardize indexers to integers than booleans. If we took that approach, I don't think it would be any more difficult to support unsorted indexing.

jhamman:
@jswhit - Supporting integer indexing would allow for other lazy operations in more cases than the one I explained above.  I can image other sorting operations that are not performed because of a malformed dataset.

jswhit:
@shoyer, you may be right - but I think the counterargument is that boolean indexing is more common so standardizing around boolean indexing is preferred.

shoyer:
To be a little more explicit, any boolean indexer can be converted into an integer indexer by calling `np.flatnonzero` on it. That's exactly how NumPy handles boolean indexers internally.

For user code, boolean indexing may indeed be more common, but for applications, I suspect integer indexing is more common, because it's more general. For example, xray will always convert boolean indexers into integer indexers before indexing on a netCDF4 variable.

jswhit:
created a new branch`` unsorted_indices`` to work on this

jswhit:
[...]

PR #469, opened 2015-10-05
allow for unsorted integer array indices, and integer indices with duplicates

jswhit:
addresses issue #467 by standardizing slicing code around integer arrays instead of boolean arrays.


netCDF4/utils.py
```diff
@@ -267,22 +258,19 @@ def _StartCountStride(elem, shape, dimensions=None, grp=None, datashape=None,\
...
-                if len(indices) >= 2:
-                    step = indices[1] - indices[0]
-                else:
-                    step = None
-                if np.array_equal(indices, np.arange(start, stop, step)):
-                    newElem.append(slice(start, stop, step))
-                else:
-                    newElem.append(ea)
+        # Replace sequence of indices with slice object if possible.
+        elif np.iterable(e) and len(e) > 1:
```
shoyer: I would consider adding another case for `len(e) == 1`, which I expect is actually somewhat common. We could replace that with `slice(e[0], e[0] + 1, 1)`.

netCDF4/utils.py
```diff
@@ -267,22 +258,19 @@ def _StartCountStride(elem, shape, dimensions=None, grp=None, datashape=None,\
...
+            start = e[0]
+            stop = e[-1]+1
+            step = e[1]-e[0]
+            try:
+                ee = range(start,stop,step)
```
shoyer: I would add a `range = xrange` at the top of this file on Python 2.  That means you don't actually need to allocate a Python list to check the length of the range.

shoyer:
look very nice, thanks!

netCDF4/utils.py
jswhit: Done.

netCDF4/utils.py
jswhit: Done.

jhamman:
Thanks @jswhit!

PR #476, opened 2015-10-19
allow non-contiguous dates for Gregorian calendar

mcuntz:
Allow non-contiguous dates for Gregorian calendar, i.e. there can be dates before and after the missing days in 1582.

Before it was checked if min(dates) and max(dates) are below or above the missing five days in 1582. Now each julian date is checked that it is not in-between these five days.

2 checks in tst_netcdftime.py fail and 78 checks fail in the test directory with run_all.py. But these these tests all fail already in the master directory https://github.com/Unidata/netcdf4-python/

Regards
Matthias

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,6 +1,7 @@
  since version 1.2.1
  ===================
  * fix bug in nc3tonc4 with --unpackshort=1 (issue 474).
+ * dates do not have to be contiguous, i.e. can be before and after the missing dates in Gregorian calendar
 
  version 1.2.1 (tag v1.2.1rel)
  =============================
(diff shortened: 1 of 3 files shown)


dopplershift:
I think it'd be useful to add a test that verifies the new behavior.

jswhit:
Agree it needs a test.  Also, no tests are failing here.

mcuntz:
Dear Jeff,

I added a test.

I updated my whole python system and now all tests work.

Regards
Matthias

jswhit:
Looks good, thanks!

PR #482, opened 2015-11-05
include cdf-5 (NC_FORMAT_64BIT_DATA) support 

jswhit:
which will be included in the netcdf-c 4.4.0 release.  Right now, the netcdf-c cdf-5 branch is needed to test this.


WardF:
If you have docker installed, you can run the tests against the appropriate branches as follows:

    $ docker run -it -e CBRANCH=cdf-5 -e PBRANCH=cdf-5 unidata/nctests:serial

Docker will download the `unidata/nctests:serial` image if it is not already installed.  The next 4.4.0 release candidate should be out tomorrow, followed (hopefully) quickly with the full release.

jswhit:
tested with the cdf-5 branch and was able to create a cdf-5 file with a uint64 variable. I think this is ready to merge into master.

PR #487, opened 2015-11-14
don't use ``NC_MAX_DIMS`` or ``NC_MAX_VARS``

jswhit:
to allocate space internally for variable and dimension ids.  In ``NETCDF4`` and ``NETCDF3_64BIT_DATA`` formatted files, these constants can be exceeded.

Note that we still use ``NC_MAX_NAME`` (= 256)  to allocate space internally for attribute and variable names, since there is apparently no way to determine the length of these names ahead of time (https://github.com/Unidata/netcdf-c/issues/153).


DennisHeimbigner:
So I took a look at the libsrc4 code. It appears that if gets a file that has
too many vars (> NC_MAX_VARS), it will throw an error.
What is interesting is that NC_MAX_DIMS is only checked in the netcdf-3 code
I can see no place in the netcdf-4 code where it tests NC_MAX_DIMS.

Issue #438, opened 2015-07-20
can't install under Ubuntu 15.04 / Debian 8


kain88-de:
I'm trying to install netcdf4 with pip and the build step fails complaining that it can't find `H5public.h`

```
x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fno-strict-aliasing -D_FORTIFY_SOURCE=2 -g -fstack-protector-strong -Wformat -Werror=format-security -fPIC -I/usr/include -I/home/malinke/.local/lib/python2.7/site-packages/numpy/core/include -I/usr/include/python2.7 -c netCDF4/_netCDF4.c -o build/temp.linux-x86_64-2.7/netCDF4/_netCDF4.o
    netCDF4/_netCDF4.c:273:22: fatal error: H5public.h: No such file or directory
     #include "H5public.h"
                          ^
```

I've installed the `libhdf5-dev` package and since I get to the compilation stage the setup.py seems to find the headers.  `dpkg` says that the headers are installed in `/usr/include/hdf5/serial`. This path it no passed correctly to the compiler which I assume is the reason it can't find the header. 

```
dpkg -L libhdf5-dev
/.
/usr
/usr/lib
/usr/lib/x86_64-linux-gnu
/usr/lib/x86_64-linux-gnu/hdf5
/usr/lib/x86_64-linux-gnu/hdf5/serial
/usr/lib/x86_64-linux-gnu/libhdf5_serial.settings
/usr/lib/x86_64-linux-gnu/libhdf5_serial_fortran.a
/usr/lib/x86_64-linux-gnu/libhdf5_cpp.a
/usr/lib/x86_64-linux-gnu/libhdf5_serialhl_fortran.a
/usr/lib/x86_64-linux-gnu/libhdf5_serial.a
/usr/lib/x86_64-linux-gnu/pkgconfig
/usr/lib/x86_64-linux-gnu/pkgconfig/hdf5-serial.pc
/usr/lib/x86_64-linux-gnu/libhdf5_hl_cpp.a
/usr/lib/x86_64-linux-gnu/libhdf5_serial_hl.a
/usr/include
/usr/include/hdf5
/usr/include/hdf5/serial
/usr/include/hdf5/serial/h5d.mod
/usr/include/hdf5/serial/H5CompType.h
/usr/include/hdf5/serial/hdf5.h
...
```

[...]

Issue #430, opened 2015-06-05
netcdf4-python does not build out of the box on a Debian (pip or from github master)


jmp75:
Hi,

I have I think all the prerequisites packages to build from source, including hdf5:
```
pkg-config --cflags hdf5
## -I/usr/include/hdf5/serial 
```
but `H5public.h` fails to be included:
```
cd ~/src/github_jm/netcdf4-python
python3 setup.py build
## snipped
##creating build/temp.linux-x86_64-3.4/netCDF4
##x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fstack-protector-strong -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -fPIC -I/usr/include -I/usr/lib/python3/dist-packages/numpy/core/include -I/usr/include/python3.4m -c netCDF4/_netCDF4.c -o build/temp.linux-x86_64-3.4/netCDF4/_netCDF4.o
netCDF4/_netCDF4.c:237:22: fatal error: H5public.h: No such file or directory
## #include "H5public.h"
##                      ^
## compilation terminated.
```

Further information below on the environment.
Happy to submit a pull request if I find a proper fix, but am unfamiliar with python package build, so may not find my way around it.
May be a shortcoming of `nc-config` rather than netcdf4-python. 

Debian 3.16.7-ckt4-1 (2015-01-16) x86_64 GNU/Linux  
(Debian "unstable")

```
$ aptitude show libnetcdf-dev
# State: installed
# Version: 1:4.1.3-7.2
```
```
$ aptitude show libhdf5-dev
# Version: 1.8.13+docs-15
# Replaces: libhdf5-serial-dev (< 1.8.12-9~), libhdf5-serial-dev (< 1.8.12-9~)
# Provides: libhdf5-serial-dev
```

jswhit:
You can edit setup.cfg and specify the locations of the hdf5 headers and libs (and turn off nc-config by setting use_ncconfig=False).

jmp75:
I will use it as a workaround on my machine, but I wonder whether this is unusual. I don't think my setup is very unusual. Then again, if it was not reported...

jswhit:
What does nc-config ``nc-config --cflags`` report?

jswhit:
Is this still a problem?

PR #518, opened 2016-01-28
Adjust setup.py to find netcdf and HDF5 on debian 8

bluesquall:
Use `pkg-config` to find appropriate compiler flags for dependencies.

Note that has been tested for pip and python3.4.2 on a Debian8 vm using:
```shell
$ pip install -vvv git+ssh://github.com/bluesquall/netcdf4-python.git@debian-pkg-config#egg=netCDF4
```
that is, without needing to manually set `USE_SETUPCFG=0`, `HDF5_INCDIR`, and `HDF5_LIBDIR` as in issues #438 and #430.

On this machine, the `netcdf-bin` package is *not* installed, so `nc-config` does not exist. On a related not, if `nc-config` were installed, it would return different values for the libraries and include directories than what `pkg-config` returns. This is probably a bug in `nc-config` upstream for Debian and variants, as noted in [this comment](https://github.com/Unidata/netcdf4-python/issues/438#issuecomment-123110407). I don't have time to dig into that deeper right now, but it's on my backburner list.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -248,7 +248,20 @@ if not retcode:
     lib_dirs = [str(l[2:].decode()) for l in dep.split() if l[0:2].decode() == '-L' ]
     dep=subprocess.Popen([ncconfig,'--cflags'],stdout=subprocess.PIPE).communicate()[0]
     inc_dirs = [str(i[2:].decode()) for i in dep.split() if i[0:2].decode() == '-I']
-# if nc-config didn't work (it won't on windows), fall back on brute force method
+elif subprocess.call(['pkg-config','--libs', 'hdf5'],stdout=subprocess.PIPE) == 0:
+    sys.stdout.write('using pkg-config ...\n')
+    dep=subprocess.Popen(['pkg-config','--libs','netcdf'],stdout=subprocess.PIPE).communicate()[0]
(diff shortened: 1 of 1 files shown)


jswhit:
Looks good - thanks!

ocefpaf:
This PR broke Windows builds that uses `setup.cfg` because `pkg-config` does not exist in that platform.  (And I believe it will have an unexpected effect on Linux and OS X trying to use `setup.cfg` too.)

See https://ci.appveyor.com/project/conda-forge/netcdf4-feedstock/build/1.0.1/job/88yf12qr9p7ud71d.

jswhit:
Thanks @ocefpaf, I've uploaded a new tarball with pypi with this fix.

How hard is it to set up an automated github test for windows with appveyor?

ocefpaf:
> How hard is it to set up an automated github test for windows with appveyor?

Medium difficulty. I should be able to do something using conda, because `hdf5` and `netcdf4` would be a pain to set up in AppVeyor without conda, would that be OK?

jswhit:
Sure, don't see why not.

ocefpaf:
PR inbound...

Issue #529, opened 2016-02-18
Round trip converts text attributes to strings


jjhelmus:
When netcdf4-python is used to read a "text" attribute from a file and write that attribute to a second file the resulting attribute will be a string where-as the original attribute was "text".

For example:

``` Python
import netCDF4

# create a reference netCDF file with a text (non-unicode) attribute
ds1 = netCDF4.Dataset('test1.nc', 'w')
ds1.setncattr('foo', 'bar')
ds1.close()

# copy the attribute to a second file, result is a string attribute
ds1 = netCDF4.Dataset('test1.nc', 'r')
ds2 = netCDF4.Dataset('test2.nc', 'w')
ds2.setncattr('foo', ds1.getncattr('foo'))
ds1.close()
ds2.close()
```

ncdumps of the two files show that the text attribute was changed to a string attribute.

```
 ncdump test1.nc; ncdump test2.nc
netcdf test1 {

// global attributes:
		:foo = "bar" ;
}
netcdf test2 {

// global attributes:
		string :foo = "bar" ;
}
```

From what I can tell this is a result of #389 which addressed #388.  Unicode Python variables are written as string attributes (using nc_put_att_string) but reading a "text" or string attribute returns unicode.  One possible fix would be to have "text" attribute be returned as str (Python 2) or bytes (Python 3).

jjhelmus:
I think the issue is that both NC_CHAR and NC_STRING attribute types are decoded to unicode on [lines 1101-1121 of _netCDF4.pyx](https://github.com/Unidata/netcdf4-python/blob/master/netCDF4/_netCDF4.pyx#L1101-L1121).  I'm not sure what would happen if NC_CHAR attributes were not decoded on line 1113 but it might be an option.

jswhit:
I don't think there is any easy way to preserve the round-trip, especially with python 3, without adding significant complication and confusion.  Under what conditions would this actually be a problem?

jjhelmus:
The issues seems to be that other netcdf software does not support the "string" type qualifier on attributes.  In the case I ran into was that one of our users was having issues reading a file in Matlab after the "string" attribute modification were made (ARM-DOE/pyart#476).  A similar issue in xarray(pydata/xarray#681) has a bit more discussion and an alternative to try encoding all strings to ascii prior to writing the file.  

@shoyer or @jhamman might be able to provide more details on on this topic.

jswhit:
[...]

PR #530, opened 2016-02-19
make sure NC_CHAR attributes returned as ascii strings (issue #529)

jswhit:
In python3, ascii attributes of type NC_CHAR will now be returned as bytes, not strings.  This will potentially break some code.


netCDF4/_netCDF4.pyx
```diff
@@ -4912,6 +4914,21 @@ cdef _read_enum(group, nc_type xtype, endian=None):
...
+    try:
+        return string.encode(encoding)
+    except UnicodeEncodeError:
+        return string
+
+def _bytes_to_str(b):
+    # if python 3 and b is a bytes object, convert bytes to ascii string.
+    # otherwise do nothing.
+    if python3 and type(b) == bytes:
+        return str(b,encoding='ascii')
```
jhamman: space after `,`

netCDF4/_netCDF4.pyx
```diff
@@ -5145,6 +5164,7 @@ datetime objects. The datetime instances
 do not contain a time-zone offset, even if the specified `units`
 contains one.
     """
+    units = _bytes_to_str(units); calendar = _bytes_to_str(calendar)
```
jhamman: I'd put these on separate lines.

jhamman:
@jswhit - two minor comments and one question?  Does this give us complete roundtrip behavior for attributes that are written as python 3 bytestrings  and unicode strings?

jswhit:
With these changes, the attribute types should be preserved on roundtrip, both for python 2 and 3.

jhamman:
Great.  This looks good to me then.

jswhit:
OK, thanks - merging now...

PR #531, opened 2016-02-25
Revert "make sure NC_CHAR attributes returned as ascii strings (issue #529)"

jswhit:
Reverts Unidata/netcdf4-python#530


jswhit:
reverting this pull request merge, since more discussion is needed on its implications.  See issue #529.

PR #533, opened 2016-02-26
try to write all attributes as NC_CHAR (issue #529)

jswhit:
only write NC_STRING if it can't be converted to ascii.


netCDF4/_netCDF4.pyx
```diff
@@ -1240,8 +1241,19 @@ cdef _set_att(grp, int varid, name, value, nc_type xtype=-99):
                     ierr = nc_del_att(grp._grpid, varid, attname)
                     if ierr != NC_NOERR:
                         raise RuntimeError((<char *>nc_strerror(ierr)).decode('ascii'))
-            # a unicode string, use put_att_string (if NETCDF4 file).
-            ierr = nc_put_att_string(grp._grpid, varid, attname, 1, &datstring)
+            # try to convert to ascii string, write as NC_CHAR 
+            # else it's a unicode string, write as NC_STRING (if NETCDF4)
+            try:
+                if force_ncstring: raise UnicodeDecodeError("",b"",1,1,"")
```
shoyer: Why do you need this strange argument here to UnicodeDecodeError?

netCDF4/_netCDF4.pyx
```diff
@@ -1240,8 +1241,19 @@ cdef _set_att(grp, int varid, name, value, nc_type xtype=-99):
...
+                if python3:
+                    str(dats,encoding='ascii')
+                else:
+                    dats = dats.encode('ascii')
```
shoyer: I think this works, but the logic is feels pretty convoluted.

test/tst_atts.py
```diff
@@ -81,10 +82,17 @@ def setUp(self):
...
         # with version 1.2.1 without pull request #486)
         f.foo = NP.array('bar','S')
         f.foo = NP.array('bar','U')
+        # issue #529 write string attribute as NC_CHAR unless
+        # it can't be decoded to ascii.  Add setncattr_string
+        # method to force NC_STRING.
+        f.charatt = u'foo' # will be written as NC_CHAR
+        f.setncattr_string('stringatt','bar') # NC_STRING
+        f.cafe = u'caf\xe9' # NC_STRING
+        v.setncattr_string('stringatt','bar') # NC_STRING
```
shoyer: Can we add an example writing bytes as an attribute? Seems like it's worth testing for completeness.

netCDF4/_netCDF4.pyx
jswhit: Because UnicodeDecodeError has 5 mandatory args 

https://coderwall.com/p/stzy9w/raising-unicodeencodeerror-and-unicodedecodeerror-manually-for-testing-purposes

that all have to be specified to raise it manually.

test/tst_atts.py
jswhit: Done

netCDF4/_netCDF4.pyx
jswhit: It's suprisingly hard to write unicode-string related code that works for python 2 and 3 that doesn't look convoluted :.(

netCDF4/_netCDF4.pyx
jswhit: Changed it to UnicodeError, which doesn't require args.

PR #540, opened 2016-03-11
AppVeyor

ocefpaf:
@jswhit this is not ready yet. We need to,

- [x] create an AppVeyor account
- [x] activate AppVeyor on this repository
- [X] clean up the copy-and-paste of the `appveyor.yml` and recipe bolierplate
- [X] run the netcdf4-python test suit

Only you can do the first two as it requires adim privileges. I will take care of the last two once AppVeyor is enable.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,43 +19,17 @@ environment:
     - TARGET_ARCH: x86
       CONDA_NPY: 110
       CONDA_PY: 27
+
     - TARGET_ARCH: x64
       CONDA_NPY: 110
       CONDA_PY: 27
 
     - TARGET_ARCH: x86
-      CONDA_NPY: 19
-      CONDA_PY: 27
-    - TARGET_ARCH: x64
-      CONDA_NPY: 19
-      CONDA_PY: 27
-
-    - TARGET_ARCH: x86
-      CONDA_NPY: 110
-      CONDA_PY: 34
-    - TARGET_ARCH: x64
-      CONDA_NPY: 110
-      CONDA_PY: 34
-
-    - TARGET_ARCH: x86
-      CONDA_NPY: 19
-      CONDA_PY: 34
-    - TARGET_ARCH: x64
-      CONDA_NPY: 19
-      CONDA_PY: 34
-
-    - TARGET_ARCH: x86
-      CONDA_NPY: 110
-      CONDA_PY: 35
-    - TARGET_ARCH: x64
       CONDA_NPY: 110
       CONDA_PY: 35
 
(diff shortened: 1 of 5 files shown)


jswhit:
OK, first two on the list are done.

ocefpaf:
@jswhit I closed and re-opened the PR to see if AppVeyor would kick-in, but nothing happened.

jswhit:
I initiated an appveyor build manually, but it's trying to run on github master.  I don't see any way to tell it to run on a pull request.

ocefpaf:
I have to confess that I set AppVeyor up only once and I don't recall all the switches to make it work.  Let me try closing and opening once more.  If that does not work I can re-send this PR.

jhamman:
You may need to push another commit to this branch.

ocefpaf:
Thanks @jhamman! Closing and opening does not work, but the extra commit did.

@jswhit even though AppVeyor is green (I did not expect that) I still want to polish this a little bit. I will ping you once I am done.

appveyor.yml
```diff
@@ -0,0 +1,88 @@
...
+    - TARGET_ARCH: x64
+      CONDA_NPY: 110
+      CONDA_PY: 35
+
+    - TARGET_ARCH: x86
+      CONDA_NPY: 19
+      CONDA_PY: 35
+    - TARGET_ARCH: x64
+      CONDA_NPY: 19
+      CONDA_PY: 35
```
ocefpaf: @jswhit AppVeyor is quite slow, so feel free to reduce this build matrix if you want. (Not sure if all that is needed.)

conda.recipe/run_test.py
```diff
@@ -0,0 +1,14 @@
+import os
+import netCDF4
+
+# Check OPeNDAP functionality.
+url = 'http://geoport-dev.whoi.edu/thredds/dodsC/estofs/atlantic'
+nc = netCDF4.Dataset(url)
+
```
ocefpaf: The first two test above are *packaging tests* and may be removed. The test below is the actual unittests for `netCDF4-python`.

jswhit:
I think we can get away with just testing np11+{x64,x86}+{py27,py35}.  I would like to keep the x86 test there, since we don't currently have any other tests for 32-bit architectures.

ocefpaf:
> I think we can get away with just testing np11+{x64,x86}+{py27,py35}. 

Makes sense.  I just pushed another commit removing the unnecessary builds.

jswhit:
OK, merging now.  Thanks @ocefpaf!

PR #543, opened 2016-03-27
Use NamedTemporaryFile instead of deprecated mktemp.

sebastic:
While investigating a test failure (caused by the absense of `ncdump`), I noticed that the tests still use the deprecated `mktemp()`, `NamedTemporaryFile()` [should be used instead](https://docs.python.org/2/library/tempfile.html#tempfile.mktemp) because it doesn't have race conditions in the file’s creation.


jswhit:
Looks good - thanks!

PR #544, opened 2016-03-30
Just a couple of typos

flamingbear:


jswhit:
The html docs are automatically generated from the docstrings in netCDF4/_netCDF4/pyx.  It would be better to fix the typos in the source code docstrings.

jswhit:
I went ahead an fixed the docstrings in _netCDF4.pyx in github master.  Thanks for catching those typos.

flamingbear:
Yeah, sorry, I figured as much, and I was looking at commits and saw some right on the html pages and didn't look any further, I'll know better next time.  Thanks.

PR #545, opened 2016-03-31
fix typo in Dataset.createVariable docstring

mtpain:
A minor thing, but it tripped me up when I referenced it.


jswhit:
great - thanks!

Issue #549, opened 2016-04-07
HDF5 1.10 leads to backwards incompatiable netCDF4 files


lesserwhirls:
As outlined in Unidata/netcdf-c#250 (and this [email](http://www.unidata.ucar.edu/mailing_lists/archives/netcdfgroup/2016/msg00061.html) message), the latest version of HDF5 (1.10) does not write backwards compatible hdf5 files, which means any users who builds netcdf-c against the latest version will produce netCDF4 files that are backwards incompatible. This will be a huge issue for data producers or anyone wishing to share their netCDF files with others. 
 
Not sure how netCDF4-Python would like to handle this, but we will certainly be telling users of netCDF-Java and the THREDDS Data Server (which call out to the netCDF-C library) to not use the HDF5 v1.10 until the situation becomes more clear. Note that HDF5 v.10 can read files produced by HDF5 v1.8.x, just not the other way. As far as I know, nothing is gained in the netCDF-C layer by using HDF v1.10, so sticking to HDF v1.8.x is probably best from a users prospective.

jswhit:
Yikes!  Perhaps the python module should issue a warning if it detects hdf5 1.10.   At the very least, we should change the build instructions.

Dave-Allured:
It appears there is a compatibility mode in HDF5 1.10 for writing downward compatible netcdf-4 files. Would someone please confirm my interpretation of the release documentation? Please read under "Obtaining, Testing, and Experimenting with This Software" on this page:

[https://www.hdfgroup.org/HDF5/docNewFeatures/](url)

Files will be written in the older format if these two conditions are met:

1.  Set write compatibility:

- H5Pset_libver_bounds (file_id, low=H5F_LIBVER_EARLIEST_F, high=H5F_LIBVER_LATEST_F)

2.  Avoid using the so-called "new features " in HDF5 1.10. In other words, do not change current HDF5 function sequences for writing Netcdf-4 files.

--Dave

WardF:
It looks like there's a straight-forward fix available as pointed out by @Dave-Allured.  I'm doing formal testing right now but will strive to have a release candidate out ASAP with the fix, with a full release to follow as soon as I can.  I'm thinking Monday sometime, but the Unidata usercomm meeting is that day so we'll see.  If possible I'll do it this weekend.

[...]

PR #552, opened 2016-04-10
print warning message if hdf5 1.10.x is being used (issue #549)

jswhit:


dopplershift:
👍 I like having this out ASAP, even in advance of the 4.4.1 netcdf version.

PR #559, opened 2016-04-16
Fix spelling error: singed -> signed

sebastic:
The lintian QA tool reported a spelling errors for the netcdf4-python 1.2.4 Debian package build.


jswhit:
good catch- thanks!

PR #542, opened 2016-03-20
AppVeyor Rolling builds

ocefpaf:
AppVeyor puts all PR updates in a test queue that can get very big (and slow) with just a few commit.
This PR cancels any previous test if a new commit exist on the same PR.

PR #560, opened 2016-04-16
Add msinttypes

ocefpaf:
Test to fix https://github.com/Unidata/netcdf4-python/pull/542#issuecomment-210722524

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,19 +17,19 @@ environment:
 
   matrix:
     - TARGET_ARCH: x86
-      CONDA_NPY: 110
+      CONDA_NPY: 111
       CONDA_PY: 27
 
     - TARGET_ARCH: x64
-      CONDA_NPY: 110
+      CONDA_NPY: 111
       CONDA_PY: 27
 
     - TARGET_ARCH: x86
-      CONDA_NPY: 110
+      CONDA_NPY: 111
       CONDA_PY: 35
 
     - TARGET_ARCH: x64
-      CONDA_NPY: 110
+      CONDA_NPY: 111
       CONDA_PY: 35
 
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -61,7 +61,7 @@ install:
     - cmd: conda update --yes --quiet conda
 
     # Conda build tools.
-    - cmd: conda install --yes --quiet obvious-ci --channel conda-forge
+    - cmd: conda install --yes --quiet obvious-ci --channel conda-forge msinttypes
```
ocefpaf: @jswhit you will notice that Python 3.5 builds are fine because the Visual Studio used for Python 3.5 does have `stdint.h`

Here I am adding the `msinttypes` packages that should provide those. Not sure if I need to set some env var to ensure that `netcdf4-python` finds it.

conda.recipe/meta.yaml
```diff
@@ -18,15 +18,16 @@ requirements:
     build:
         - python
         - setuptools
-        - numpy x.x
         - cython
-        - hdf5
+        - numpy x.x
+        - msinttypes  # [win and py<35]
```
ocefpaf: @jswhit you will notice that Python 3.5 builds are fine because the Visual Studio used for Python 3.5 does have `stdint.h`

Here I am adding the `msinttypes` packages that should provide those. Not sure if I need to set some env var to ensure that `netcdf4-python` finds it.

ocefpaf:
@jswhit I updated the recipe to use the latest numpy and pinned the `hdf5` version down to avoid a mismatch with `libnetcdf`.

The `msinttypes` package brings the `stdint.h` and AppVeyor should be green again.

PR #564, opened 2016-05-29
Remove duplicate verbs defines creates

kinow:
Hi, 

I believe where it reads "A Dataset creates defines creates a special group", it was intended to be "A Dataset creates a special group".

Thanks
Bruno


jswhit:
Thanks!

PR #567, opened 2016-06-21
Correct Documentation Parenthesis Position

jeixav:
I think the position of a parenthesis in the documentation is wrong:
> Compound data types map directly to numpy structured (a.k.a 'record' arrays).

If the part in parentheses were removed, the sentence would be incomplete, i.e.
> Compound data types map directly to numpy structured.

This pull request moves the position of the parenthesis to correct the problem:
> Compound data types map directly to numpy structured (a.k.a 'record') arrays.


jswhit:
Please don't modify the html directly, it's automatically generated from the docstrings.  _netCDF4.pyx is the place to make the fix.  Thanks!

jeixav:
Thank you for the feedback @jswhit. I think I changed the right file in my last commit; let me know if not.

jswhit:
Great - thanks!

Issue #570, opened 2016-07-06
MFDataset.set_auto_maskandscale fails


j08lue:
The issue: `set_auto_maskandscale` and the like seem not to work on a `MFDataset`.

```python
import netCDF4
ds = netCDF4.MFDataset('some/glob/pattern.nc')
ds.set_auto_maskandscale(True)
```

thows

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-818578dc39d8> in <module>()
----> 1 ds.set_auto_maskandscale(True)

netCDF4\_netCDF4.pyx in netCDF4._netCDF4.Dataset.set_auto_maskandscale (netCDF4\_netCDF4.c:20136)()

AttributeError: 'NoneType' object has no attribute 'values'
```

For sample data, see

    ftp://ftp.cdc.noaa.gov/Datasets/ncep.reanalysis/surface/air.sig995.2015.nc
    ftp://ftp.cdc.noaa.gov/Datasets/ncep.reanalysis/surface/air.sig995.2016.nc

This is version 1.2.4

jswhit:
works for individual variables, but not on the MFDataset instance.  Should be easy to fix, watch this space.

jswhit:
see pull request #571

jswhit:
pull request #571 merged, closing now...

PR #571, opened 2016-07-06
fix for issue #570 (missing MFDataset.set_auto_maskandscale)

jswhit:
also added set_auto_scale and set_auto_mask.


j08lue:
Great, thanks @jswhit!

PR #573, opened 2016-07-26
fixed docstring for keepweakref parameter

ChrisBarker-NOAA:
The docstring had `weakref=False`, rather than `keepweakref=False`, which means a user could not get the functionality desired.

I also added a bit more to the docs about why one would want to use weak references.

### DIFF
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -1679,7 +1679,7 @@ references to the parent Dataset or Group.
                  diskless=False, persist=False, keepweakref=False, **kwargs):
         """
         **`__init__(self, filename, mode="r", clobber=True, diskless=False,
-        persist=False, weakref=False, format='NETCDF4')`**
+        persist=False, keepweakref=False, format='NETCDF4')`**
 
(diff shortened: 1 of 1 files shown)


jswhit:
Great - thanks Chris.  Merging now.

ChrisBarker-NOAA:
that was fast -- thanks!

BTW -- we're running into issues with too many files open in an app that, well opens a lot of files. But it seems they are not getting cleaned up properly. Which I expected due to the circular reference issues (hence my looking into the `keepweakref` parameter).

However, I just did a simple experiment, and it looks like now, with recent python2.7 and netCDF4, the gc CAN, in fact, clean up `netcdf4.Dataset` objects.

Did support for Cyclic Garbage Collection get implemented?

https://docs.python.org/2/c-api/gcsupport.html

-Chris

jswhit:
Nothing was added in netcdf4-python, but perhaps cython's support for cyclic garbage collection has improved.

ChrisBarker-NOAA:
that may be it -- I do see `Py_TPFLAGS_HAVE_GC` and `tp_traverse` in the generated C code.

Very cool -- Cython really is great!

Issue #576, opened 2016-07-29
Check valid_min, valid_max and valid_range when masking


duncanwp:
It would be great if the masking would respect valid_min, max and range when loading variables - to be honest I was a little surprised it doesn't already. 

This is fairly easy to do manually, but because it has to come before the scaling it means I have to do the masking _and_ scaling myself which is a bit of a pain.

jswhit:
At 

http://www.unidata.ucar.edu/software/netcdf/docs/attribute_conventions.html

I see

"Generic applications should treat values outside the valid range as missing"  

Are you suggesting that if a value is outside the valid range, it should be masked (i.e. treated the same way as a missing value)?    I wonder what other applications do.

duncanwp:
Precisely. Those conventions also define valid_min and valid_max as well.

I've implemented this recently in https://github.com/cedadev/cis due to user feedback, here: https://github.com/cedadev/cis/blob/master/cis/data_io/netcdf.py#L252, but I'm sure there are more elegant solutions. Particularly - the conventions state that "valid_range must not be defined if either valid_min or valid_max is defined", but I don't enforce that. I also only warn if I can't parse the metadata as I don't consider it a show-stopper, but you might want to be stricter.

jswhit:
created a branch (valid_minmax) to work on this.

jswhit:
implemented this part:

"If neither valid_min, valid_max nor valid_range is defined then generic applications should define a valid range as follows. If the data type is byte and _FillValue is not explicitly defined, then the valid range should include all possible values. Otherwise, the valid range should exclude the _FillValue (whether defined explicitly or by default) as follows. If the _FillValue is positive then it defines a valid maximum, otherwise it defines a valid minimum. "

jswhit:
@duncanwp, will wait for your input before merging.

duncanwp:
Wow, that was quick, thanks! I've made a few comments inline.

jswhit:
The pull request was merged, closing now...

duncanwp:
Great, thanks @jswhit

PR #577, opened 2016-08-02
use valid_min/valid_max/valid_range when creating mask (issue #576)

jswhit:


netCDF4/_netCDF4.pyx
```diff
@@ -3754,27 +3757,64 @@ rename a `netCDF4.Variable` attribute named `oldname` to `newname`."""
...
-                mask = hasmval
-            else:
-                mask = None
-            if mask is not None:
-                fill_value = mval
-                totalmask += mask
+                if mvalisnan: 
+                    mvalmask += numpy.isnan(data)
+                else:
+                    mvalmask += data==mval
```
duncanwp: According to the conventions the missing value "... attribute is not treated in any special way by the library or conforming generic applications, but is often useful documentation and may be used by specific applications. The missing_value attribute can be a scalar or vector containing values indicating missing data. These values should all be outside the valid range so that generic applications will treat them as missing."

I for one would expect missing_values in the data to be masked whether inside the valid range or not - but that's not what the conventions appear to be saying. What's your take on this?

netCDF4/_netCDF4.pyx
```diff
@@ -3754,27 +3757,64 @@ rename a `netCDF4.Variable` attribute named `oldname` to `newname`."""
...
+            # treatment of byte data as described at
+            # http://www.unidata.ucar.edu/software/netcdf/docs/attribute_conventions.html).
+            if hasattr(self, 'valid_range') and len(self.valid_range) == 2:
+                validmin = numpy.array(self.valid_range[0], self.dtype)
+                validmax = numpy.array(self.valid_range[1], self.dtype)
+            else:
+                if hasattr(self, 'valid_min'):
+                    validmin = numpy.array(self.valid_min, self.dtype)
+                if hasattr(self, 'valid_max'):
+                    validmax = numpy.array(self.valid_max, self.dtype)
```
duncanwp: Is it worth trying to catch invalid dtypes for these attributes, or are you happy for any errors to bubble up?

netCDF4/_netCDF4.pyx
```diff
@@ -3754,27 +3757,64 @@ rename a `netCDF4.Variable` attribute named `oldname` to `newname`."""
...
+            # (whether defined explicitly or by default) as follows. 
+            # If the _FillValue is positive then it defines a valid maximum,
+            #  otherwise it defines a valid minimum."
+            byte_type = self.dtype.str[1:] in ['u1','i1']
+            if hasattr(self, '_FillValue'):
+                fval = numpy.array(self._FillValue, self.dtype)
+            else:
+                fval = numpy.array(default_fillvals[self.dtype.str[1:]],self.dtype)
+                if byte_type: fval = 0.
+            if validmin is None and fval < 0:
```
duncanwp: I like this logic, and it does conform to the spec, but I wonder what happens if a _FillValue is defined as 0? It's obvously a bit obscure, but not inconcevable. My reading of it is that a _FillValue of 0 should be a validmin. Perhaps default fval to None?

netCDF4/_netCDF4.pyx
jswhit: I would too - in fact, that's what the code here will do.  I had not noticed that clause in the docs.

netCDF4/_netCDF4.pyx
jswhit: It tries to cast the attributes to the data type of the variable.  This could potentially fail, and it might be a good idea to catch this error.

netCDF4/_netCDF4.pyx
jswhit: Good catch - I'll fix that

Issue #593, opened 2016-10-07
netcdftime.datetime: implement __add__, __radd__, __sub__


ckhroulev:
I propose adding `__add__`, `__radd__`, `__sub__` implementations to `netcdftime.datetime`. This will make is possible to do things like

``` python
import netCDF4, datetime
print(netCDF4.netcdftime.datetime(-5000, 1, 1, 12) + datetime.timedelta(seconds=0))
```

This is prompted by #584 (the return value of `netcdftime.utime.num2date` should always work with `datetime.timedelta`).

Let me know if I should go ahead with this and submit a pull request.

The implementation I have in mind would use `utime.date2num`, shift the Julian day by `timedelta.total_seconds / 86400` and `utime.num2date` to get a date back. This would require extending `netcdftime.datetime` in `netcdftime.py` instead of using Cython (which would lead to a circular dependency).

By the way: why is `netcdftime.datetime` written in Cython?

shoyer:
We wrote this type in Cython because it was an easy way to get significant performance improvements. 

I'm all for adding more functionality to the netcdftime.datetime class, but this should not come at the price of performance. Fortunately, it's quite easy to move Python code into Cython modules, so this shouldn't be an issue.

davidhassell:
I'd also like to request that the `replace` method be implemented too. Thanks!

ckhroulev:
@shoyer Hmm. Right. I would need to come up with a way to test for performance degradation then. 

Do you know of a good way to measure the performance of `netcdftime.datetime`?

Do you recall which `netcdftime.datetime` method benefitted from using Cython?

Would the time reported by `python tst_netcdftime.py` be an adequate measure, in your opinion?

@davidhassell Sure!

shoyer:
@ckhroulev See https://github.com/Unidata/netcdf4-python/pull/303 for the original PR porting `netcdftime.datetime` to Cython and [here](https://github.com/Unidata/netcdf4-python/pull/303/files/7ad575e01c2addb945c8b1d278a441edeaac8c96#r20033706) for some benchmarks. It made date2num about 2x faster.

I'm still confused why you think "This would require extending netcdftime.datetime in netcdftime.py instead of using Cython (which would lead to a circular dependency)."

Instead, you could move date2num/num2date into the Cython module. The code would not even need to change, since Cython supports a strict superset of valid Python.

ckhroulev:
[...]

PR #594, opened 2016-10-17
Implement datetime and timedelta arithmetic. See #593.

ckhroulev:


netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,209 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+        return (self.year, self.month, self.day, self.hour,
+                self.minute, self.second, self.dayofwk, self.dayofyr, -1)
+
+    def _to_real_datetime(self):
+        return real_datetime(self.year, self.month, self.day,
+                             self.hour, self.minute, self.second,
+                             self.microsecond)
+
+    def __repr__(self):
+        return self.strftime(self.format)
```
shoyer: maybe better to change the repr to include full type information, including the calendar?

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,209 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+
+    def __hash__(self):
+        try:
+            d = self._to_real_datetime()
+        except ValueError:
+            return hash(self.timetuple())
+        return hash(d)
+
+    def __richcmp__(self, other, int op):
+        if hasattr(other, 'strftime'):
```
shoyer: Should require that calendars match (or `return NotImplemented`).

Also, should check microseconds.

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,209 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+        return (self.__class__,args)
+
+    @property
+    def calendar(self):
+        return self._calendar
+
+    @property
+    def utime(self):
+        if self._utime is None:
+            self._utime = utime("seconds since 1-1-1", self._calendar)
```
shoyer: Maybe better to make a global cache of utime instances for each calendar?

I am concerned that caching could add significant overhead per datetime object, and also it's probably common to do lots of these operations on different datetime objects.

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,209 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+        return self._utime
+
+    def __add__(self, other):
+        if isinstance(self, datetime) and isinstance(other, timedelta):
+            date = self
+            delta = other
+            utime = date.utime
```
shoyer: this shadows the global `utime` class, which is generally poor style

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,209 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+        elif isinstance(self, timedelta) and isinstance(other, datetime):
+            date = other
+            delta = self
+            utime = date.utime
+        else:
+            raise TypeError("incompatible netcdftime.datetime.__add__() arguments")
```
shoyer: Just write `return NotImplemented` here. Python takes care of turning that into a `TypeError`.

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,209 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+        return utime.num2date(utime.date2num(date) + delta.total_seconds())
+
+    def __sub__(self, other):
+        if isinstance(self, datetime): # left arg is a datetime instance
+            utime = self.utime
+            if isinstance(other, datetime):
+                # datetime - datetime
+                return timedelta(seconds=utime.date2num(self) - utime.date2num(other))
```
shoyer: Probably needs to verify that calendars match

test/tst_netcdftime.py
```diff
@@ -699,5 +699,70 @@ def test_dayofwk(self):
 
             old_date = date
 
+class issue593TestCase(unittest.TestCase):
```
shoyer: Better to use a sensible name than a meaningless issue number, unless it's for a specific regression test.

test/tst_netcdftime.py
```diff
@@ -699,5 +699,70 @@ def test_dayofwk(self):
...
+    def setUp(self):
+        self.date1 = datetimex(-5000, 1, 2, 12, calendar="365_day")
+        self.date2 = datetimex(-5000, 1, 3, 12, calendar="365_day")
+        self.real_date1 = datetimex(1969,  7, 20, 12, calendar="standard")
+        self.real_date2 = datetime(1969,  7, 21, 12)
+        self.delta = timedelta(hours=25)
+
+    def test_add(self):
+        next_day = self.date1 + self.delta
+        self.assertEqual(next_day.day ,self.date1.day + 1)
```
shoyer: test equality for the full `datetime` objects, not just the `day` field

test/tst_netcdftime.py
```diff
@@ -699,5 +699,70 @@ def test_dayofwk(self):
...
+
+        next_day = self.delta + self.date1
+        self.assertEqual(next_day.day ,self.date1.day + 1)
+
+        with self.assertRaises(TypeError):
+            print(self.date1 + 1)
```
shoyer: No need to include `print()` here

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,209 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+            self_str = self.strftime('%Y-%m-%d %H:%M:%S')
+            other_str = other.strftime('%Y-%m-%d %H:%M:%S')
+            return PyObject_RichCompare(self_str, other_str, op)
+        return NotImplemented
+
+    def __reduce__(self):
+        """special method that allows instance to be pickled"""
+        args = (self.year,self.month,self.day,self.hour,
+                self.minute,self.second,self.microsecond,
+                self.dayofwk,self.dayofyr)
```
shoyer: needs to be updated to include calendar.

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,209 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+of the string produced by __repr__ is controlled by self.format
+(default %Y-%m-%d %H:%M:%S). Does support comparisons with other
+phony datetime and with datetime.datetime objects.
+
+Instance variables are year,month,day,hour,minute,second,dayofwk,dayofyr
+and format.
+    """
+    cdef readonly int year, month, day, hour, minute, dayofwk, dayofyr
+    cdef readonly int second, microsecond
+    cdef str _calendar
```
shoyer: should be `readonly` (then you don't need a separate property)

test/tst_netcdftime.py
```diff
@@ -699,5 +699,70 @@ def test_dayofwk(self):
...
+        # subtracting netcdftime.datetime from datetime.datetime
+        delta = self.real_date2 - self.real_date1
+        # real_date2 and real_date1 are exactly one day apart
+        self.assertEqual(delta.total_seconds(), 86400)
+
+        # subtracting datetime.datetime from netcdftime.datetime
+        delta = self.real_date1 - self.real_date2
+        # real_date2 and real_date1 are exactly one day apart
+        self.assertEqual(delta.total_seconds(), -86400)
+
```
shoyer: needs a test with different calendars

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,209 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+            elif isinstance(other, real_datetime):
+                # datetime - real_datetime
+                if self.calendar not in ("standard", "gregorian"):
+                    raise ValueError("cannot compute the time difference between dates with different calendars")
+                return self._to_real_datetime() - other
+            elif isinstance(other, timedelta):
+                # datetime - timedelta
+                return utime.num2date(utime.date2num(self) - other.total_seconds())
+            else:
+                raise TypeError("incompatible netcdftime.datetime.__sub__() arguments")
```
shoyer: you can just return NotImplemented

davidhassell:
Hello,

Should `datetime.__reduce__` include the `calendar` attribute?

shoyer:
> Should datetime.__reduce__ include the calendar attribute?

Yes. This is necessary for pickle to work properly. If there isn't already a pickle test (looks like it's missing), please add one -- should be as simple as `self.assertEqual(dt, pickle.loads(pickle.dumps(dt))`.

jswhit:
There appears to be some python 2.6 failures in Travis.  `timedelta.total_seconds()` debuted in python 2.7.   I note in the docs 

Equivalent to (td.microseconds + (td.seconds + td.days * 24 * 3600) * 10**6) / 10**6 computed with true division enabled.

ckhroulev:
@jswhit Thanks -- I'm aware of this.

I am in the process of cleaning up a different implementation that does not use `timedelta.total_seconds()`. I'll push it soon (today).

netcdftime/_netcdftime.pyx
ckhroulev: Right. Thanks! (It is possible to use two different utime instances and support arguments with different calendars, but we can do it later, if necessary).

netcdftime/_netcdftime.pyx
ckhroulev: This is taken care of by moving the cache of utime instances out of `datetime`.

netcdftime/_netcdftime.pyx
ckhroulev: Done. (I'll post a before-and-after timing comparison in a pull request comment.)

netcdftime/_netcdftime.pyx
ckhroulev: Yes. (I also added a regression test checking reversibility of pickling.)

netcdftime/_netcdftime.pyx
ckhroulev: I replaced the old `__repr__` implementation with the one including the full state of an instance. The old `__repr__` is more of a `__str__`, really.

netcdftime/_netcdftime.pyx
ckhroulev: Done. Thanks!

test/tst_netcdftime.py
ckhroulev: Right. I added tests to achieve 100% line coverage (verified using `coverage.py`). I'm happy to add more tests, though.

test/tst_netcdftime.py
ckhroulev: OK. Thanks!

test/tst_netcdftime.py
ckhroulev: Done.

test/tst_netcdftime.py
ckhroulev: Yep. Sorry about that.

ckhroulev:
Regarding the performance of `date2num` using the old and the new `netcdftime.datetime` implementations.

Setup:

``` python
from netCDF4 import netcdftime

import numpy as np

u_julian = netcdftime.utime("days since 1900-01-01", calendar="julian")

dates_julian = u_julian.num2date(np.arange(300000))

u_gregorian = netcdftime.utime("days since 1900-01-01", calendar="gregorian")

dates_gregorian = u_gregorian.num2date(np.arange(300000))
```

Old code:

```
In [7]: netcdftime.datetime??
File: ~/github/ckhroulev/netcdf4-python/netcdftime/_datetime.so
Type: type

In [8]: %time u_julian.date2num(dates_gregorian) 
CPU times: user 591 ms, sys: 16.2 ms, total: 607 ms
Wall time: 605 ms
Out[8]: 
array([  0.00000000e+00,   1.00000000e+00,   2.00000000e+00, ...,
         3.00004000e+05,   3.00005000e+05,   3.00006000e+05])

In [9]: %time u_gregorian.date2num(dates_gregorian)
CPU times: user 604 ms, sys: 29.6 ms, total: 634 ms
Wall time: 627 ms
Out[9]: 
array([  0.00000000e+00,   1.00000000e+00,   2.00000000e+00, ...,
         2.99997000e+05,   2.99998000e+05,   2.99999000e+05])
```

New code:

```
In [2]: %time u_julian.date2num(dates_gregorian)
CPU times: user 467 ms, sys: 13.7 ms, total: 481 ms
Wall time: 480 ms
Out[2]: 
array([  0.00000000e+00,   1.00000000e+00,   2.00000000e+00, ...,
         3.00004000e+05,   3.00005000e+05,   3.00006000e+05])

In [3]: %time u_gregorian.date2num(dates_gregorian)
CPU times: user 474 ms, sys: 17.5 ms, total: 492 ms
Wall time: 491 ms
Out[3]: 
array([  0.00000000e+00,   1.00000000e+00,   2.00000000e+00, ...,
         2.99997000e+05,   2.99998000e+05,   2.99999000e+05])

In [4]: netcdftime.datetime??
File: ~/github/ckhroulev/netcdf4-python/netcdftime/_netcdftime.so
Type: type
```

The numbers themselves are not important; all I'm trying to say is that as far as I can tell the new code is *not* any slower.

test/tst_netcdftime.py
```diff
@@ -246,7 +246,7 @@ def runTest(self):
         d2 = datetime(1995, 1, 1)
         d3 = datetimex(2001, 2, 30)
         assert hash(d1) == hash(d1)
-        assert hash(d1) == hash(d2)
+        assert hash(d1) != hash(d2) # represent the same date, but are *not* the same object
```
shoyer: This was intentional. Otherwise, you can't use these date times in a dict.

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,470 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
 
 
-def _toscalar(a):
+cdef _toscalar(a):
     if a.shape in [(), (1,)]:
         return a.item()
     else:
         return a
+
+cdef to_tuple(dt):
```
ckhroulev: The type is *not* specified here to make it possible to compare `netcdftime.datetime` to `datetime.datetime` by converting both of them to tuples. I could write a separate version exclusively for `netcdftime.datetime`, though... (Possibly implemented as a method of `netcdftime.datetime`.)

test/tst_netcdftime.py
shoyer: See https://docs.python.org/2/reference/datamodel.html#object.__hash__

"The only required property is that objects which compare equal have the same hash value"

shoyer:
This might also be a good time to consider dropping Python 2.6 support. Not even NumPy is supporting it in new releases.

test/tst_netcdftime.py
ckhroulev: OK. Thanks for pointing this out! I will revert that change.

netcdftime/_netcdftime.pyx
```diff
@@ -1151,8 +1156,470 @@ def time2index(times, nctime, calendar=None, select='exact'):
...
+    # Python's datetime.datetime uses the proleptic Gregorian
+    # calendar. This boolean is used to decide whether a
+    # netcdftime.datetime instance is comparambe to datetime.datetime.
+    cdef readonly bint calendar_is_gregorian
+
+    # The pointer to a function used by __add__() and __sub() for
+    # datetime + timedelta computations. This allows us to check the
+    # calendar once (in __init__()) instead of doing it every time
+    # __add__() and __sub__ are called.
+    cdef datetime (*_add_timedelta)(datetime, object)
```
ckhroulev: OK. Please see the new version, now using inheritance instead of a function pointer.

jswhit:
This looks good, thank you @ckhroulev.  I will wait another day or two for others to comment and then merge.

jswhit:
merging now..

PR #595, opened 2016-10-21
BUG: Increase buffer used to store filepath

### LINKED ISSUE #585: Possible memory leak
I'm getting consistent seg-faults somewhere out of netCDF-python/netCDF-c/HDF and I'm struggling to track it down. 

The only way I can reliable reproduce it is on OS X with a specific set of input data files, and a specific output filename... It also occurs on Windows but less reliably. Predictably I've been unable to reproduce it on Linux with a debug enabled build.

I've attached the [stack](https://github.com/Unidata/netcdf4-python/files/476302/stack.txt) out of OSX when it occurs, and also some valgrind [output](https://github.com/Unidata/netcdf4-python/files/476303/valgrind.txt) from the Linux test (which doesn't seg fault but does produce potential memory leaks). The most interesting lines in the valgrind output are the "Uninitialised value was created by a heap allocation" warnings starting around line 560, and then the "Invalid write of size 1" at 660, and "Address 0x600e530 is 0 bytes after a block of size 16 alloc'd" at 673. 

It's also possible these last two might be related to #506.

I'm using HDF 1.8.17, netCDF 4.4.0, netcdf-python 1.2.4.

Apologies I've not given you much to go on, if it helps I can share a recipe for the environment I'm using on OSX and the input data files (about 50Mb).

jjhelmus:
Increase the buffer used to store the filepath by one character to accommodate
the NULL character copied by the nc_inq_path function.

closes #585


jjhelmus:
Note sure if it is typical to regenerate the _netCDF4.c file in the pull request or not.  Happy to if it is desired.

dopplershift:
Awesome, thanks for running this down!

jswhit:
Please add a Changelog entry and regenerate `_netCDF.c` with all values set to zero in include/constants.pyx

jjhelmus:
Added to the Changelog and regenerated `_netCDF.c`.  Not sure how useful the latter is given that the change is excluded from that file.

jswhit:
The change will be included in _netCDF.c, if you regenerated it from the updated netCDF4.pyx.

jjhelmus:
The change to `_netCDF.pyx` made in this PR are inside the `IF HAS_NC_INQ_PATH` block and therefore, unless I am misunderstanding the instructions, are excluded from the `_netCDF.c` file when it is generated with a include/constants.pyx file with all values set to zero.

Specifically the block in question is as follows:

``` Cython
    def filepath(self):
        """
**`filepath(self)`**

Get the file system path (or the opendap URL) which was used to
open/create the Dataset. Requires netcdf >= 4.1.2"""
        cdef int ierr
        cdef size_t pathlen
        cdef char *c_path
        IF HAS_NC_INQ_PATH:
            with nogil:
                ierr = nc_inq_path(self._grpid, &pathlen, NULL)
            if ierr != NC_NOERR:
                raise RuntimeError((<char *>nc_strerror(ierr)).decode('ascii'))
            c_path = <char *>malloc(sizeof(char) * (pathlen + 1))
            if not c_path:
                raise MemoryError()
            try:
                with nogil:
                    ierr = nc_inq_path(self._grpid, &pathlen, c_path)
                if ierr != NC_NOERR:
                    raise RuntimeError((<char *>nc_strerror(ierr)).decode('ascii'))
                py_path = c_path[:pathlen] # makes a copy of pathlen bytes from c_string
            finally:
                free(c_path)
            return py_path.decode('ascii')
        ELSE:
            msg = """
filepath method not enabled.  To enable, install Cython, make sure you have
version 4.1.2 or higher of the netcdf C lib, and rebuild netcdf4-python."""
            raise ValueError(msg)
```

jswhit:
Yes, you are correct.  The _netCDF4.c file should be unchanged.  Going ahead with the merge now...

Thanks again!

netCDF4/_netCDF4.c
```diff
@@ -2054,7 +2054,7 @@ static const char __pyx_k_Class_providing_an_interface_to[] = "\nClass providing
 static const char __pyx_k_FillValue_attribute_must_be_set[] = "_FillValue attribute must be set when variable is ";
 static const char __pyx_k_MFDataset_set_auto_maskandscale[] = "MFDataset.set_auto_maskandscale";
 static const char __pyx_k_The_vltypes_dictionary_maps_the[] = "The `vltypes` dictionary maps the names of\n    variable-length types defined for the `netCDF4.Group` or `netCDF4.Dataset` to instances of the\n    `netCDF4.VLType` class.";
-static const char __pyx_k_Users_jswhit_python_netcdf4_pyt[] = "/Users/jswhit/python/netcdf4-python/netCDF4/_netCDF4.pyx";
```
j08lue: Just curious, what is all this stuff about @jjhelmus? Constants pointing to files on your machine... Why does Cython do that?

Issue #592, opened 2016-10-07
Creating attributes with arrays of strings


cmarquardt:
Is it possible to create attributes which are arrays of strings? For example,

``` python
import netCDF4 as netcdf
nc = netcdf.Dataset("seppel.nc", "w")
nc.setncattr_string("attr", ["oink", "bleek"])
nc.close()
exit()
```

Right now (python 2.7, netCDF4 1.2.4, libnetcdf 4.3.3.1), ncdump shows that rather than an array of strings, the generated attribute is a single string with a textual representation of the list:

```
netcdf seppel {

// global attributes:
            string :attr = "[\'oink\', \'bleek\']" ;
}
```

Thanks - Christian.

jswhit:
Right now you can't create attributes that are arrays of strings.  netcdf does not have a fixed length string array type, so it would have to be done with vlen strings.

cmarquardt:
Yes, that's what I had in mind - creating an arrays of vlen strings. Sorry, the example was a bit misleading as it had two strings of the same length. 

Would this be a useful feature request? To be honest, I haven't even tried if arrays of vlen strings can be used as an attribute with the netcdf-C; I think it can be done with Java...

jswhit:
resolved with pull request #597

PR #597, opened 2016-10-28
Support reading and writing string array attributes. See #592.

ckhroulev:
This adds support for reading and writing string arrays.

``` python
import netCDF4
d = netCDF4.Dataset("foo.nc", "w")
d.char_attr = ["one", "string"] # no change in behavior: writes an NC_CHAR attribute, concatenating strings
d.setncattr_string("string_attr", ["two", "strings"]) # writes an NC_STRING attribute with 2 entries
d.close()
```

Results:

```
netcdf foo {

// global attributes:
        :char_attr = "onestring" ;
        string :string_attr = "two", "strings" ;
}
```

As a side effect

```
d.setncattr_string("string_attr", 1)
```

no longer writes the string `"1"` to the global attribute `"string_attr"`. (Calling `setncattr_string` with a value that is neither a string nor a sequence of strings should probably be an error.)

_Reading_ string array attributes is gives lists of strings.


jswhit:
thanks for this @ckhroulev.  I will try this out this weekend.

jswhit:
This looks great!  I think all it needs is a Changelog entry,  an update to the `setncattr_string` docstring, and a test.

ckhroulev:
I can reproduce this test failure and will push a fix soon.

PS: It's strange that AppVeyor did not catch the failure (it did fail, though).

jswhit:
minor typo in docstring:   "attirbute" should be "attribute".

jswhit:
also, Travis tests are failing for Python 3

ckhroulev:
@jswhit Yep. I'm working on it.

jswhit:
great!  I'll wait a day or two and then merge.

jswhit:
merging now...

Issue #596, opened 2016-10-25
bug: num2date(date2num(date)) roundtrip changes dates with year < 0


ckhroulev:
Run this to reproduce:

``` python
import netcdftime

u = netcdftime.utime("seconds since 1-1-1", "julian")

date1 = netcdftime.datetime(-1, 1, 1)
date2 = u.num2date(u.date2num(date1))

print(date1)
print(date2)
```

`date1` should be equal to `date2`, but this script prints

```
  -1-01-01 00:00:00
  -2-01-01 00:00:00
```

(Tested using commit 8773adbcd324ae0cf2ba597d1d6d6b27bb28e649.)

ckhroulev:
This is one of the reasons why I decided not to use `date2num` and `num2date` to implement `datetime.__add__()` and `datetime.__sub__()`.

ckhroulev:
A related issue:

This code
``` python
import netcdftime
u = netcdftime.utime("seconds since 1-1-1", "proleptic_gregorian")
print(u.num2date(u.date2num(netcdftime.datetime(-1, 1, 1))))
```
attempts to create a `datetime.datetime` instance with the year of -1, which is not supported by `datetime`.

Here `num2date(date2num(date))` raises `ValueError` instead of evaluating to `date`.

(Tested using commit b3f12f8027ab898e7d5090db362f0783c4fa7980.)

jswhit:
I guess this is not related to pull request #586, since it happens with `proleptic_gregorian`.   I guess `date2num` should raise an error if a negative year is supplied and the calendar is not `360_day` or `365_day`?

jswhit:
Ignore my previous comment. I think the problem is that there is no year 0 in the Julian calendar, and date2num does not know this.

    import netcdftime
    from netCDF4 import date2num, num2date
    units = "days since 1-1-1"
    date1 = netcdftime.datetime(-1, 1, 1)
    t = date2num(date1,units)
    print t

gives

    -731

instead of 

    -366

jswhit:
fix is in pull request #598

jswhit:
I noticed that with the original example

    import netcdftime
    u = netcdftime.utime("seconds since 1-1-1", "julian")
    date1 = netcdftime.datetime(-1, 1, 1)
    date2 = u.num2date(u.date2num(date1))
    print date1,type(date1)
    print date2,type(date2)

now gives

[...]

PR #603, opened 2016-11-07
fix for neg years in utime.num2date (issue #596)

jswhit:


jswhit:
what about year 0?  It's not clear to me whether it should be allowed or not.

Issue #527, opened 2016-02-11
accessing unset entries in VLEN variable causes crash


jhprinz:
I created a VLEN variable for numpy.float64 like this with the first (and only) dimension unlimited.
```
<type 'netCDF4._netCDF4.Variable'>
vlen test(mydim1)
    var_type: numpy.float64
vlen data type: float64
unlimited dimensions: mydim1
current shape = (3,)
```
Currently the dimension `mydim` has length 3 although I have only set values for entries 0 and 1 like
```py
var = ncfile.variables['test']

var[0] = np.array([1.0, ...])
var[1] = np.array([0.1, ...])
```

when I try to load the content from `var[3]` this gives me the expected `KeyError` since there are only 3 and not 4 elements in the variable stored. However doing this

```py
print var[2]

>>> Crash
```

will crash my ipython kernel. The same is true for slicing

```py
print var[:]

>>> Crash
```

I am not sure what the intended behavior is for reading unset values in VLEN variables and if this is also a problem in the wrapped netCDF library.

jswhit:
You definitely shouldn't get a crash.  Does this simple example work for you?

     from netCDF4 import Dataset
    import numpy as np
    nc = Dataset('test.nc','w')
    vlen_type = nc.createVLType(np.int,'vltest')
    nc.createDimension('x',5)
    v = nc.createVariable('vl',vlen_type,'x')
    v[0]=np.arange(2,dtype=np.int)
    v[1]=np.arange(3,dtype=np.int)
    print v[2]
    print v[:]
    nc.close()

You should get

    []
    [array([0, 1]) array([0, 1, 2]) array([], dtype=int64)
     array([], dtype=int64) array([], dtype=int64)]

jhprinz:
Yes, thanks. This works fine. I played around a little to find a simple example that fails.

This does not work. I created two variables using the same dimension. First the fixed length was
used to increase the current length to 3 and then add the VLEN variable. I am using the conda to install netcdf. Current version is
```
netcdf4                   1.2.2               np110py27_0    defaults
```

netcdf 

```py
from netCDF4 import Dataset
import numpy as np
nc = Dataset('test.nc','w')
vlen_type = nc.createVLType(np.float64,'vltest')
nc.createDimension('x', None)
v = nc.createVariable('vl', vlen_type, 'x')
w = nc.createVariable('vl2', np.float64, 'x')
w[0:3] = np.arange(3,dtype=np.float64)
[...]

PR #605, opened 2016-11-16
Initialize members of vldata to make sure that nc_free_vlens() succeeds. See #527.

ckhroulev:
Note that `free(NULL)` is a no-op.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -24,6 +24,7 @@
  * Make sure negative years work in utime.num2date (issue #596).
  * raise NotImplementedError when trying to pickle Dataset, Variable,
    CompoundType, VLType, EnumType and MFDataset (issue #602).
+ * Fix for issue #527: initialize vldata[i].p in Variable._get(...).
 
  version 1.2.4 (tag v1.2.4rel)
 ==============================
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -4498,6 +4498,9 @@ The default value of `mask` is `True`
                 # regular vlen
                 # allocate struct array to hold vlen data.
                 vldata = <nc_vlen_t *>malloc(totelem*sizeof(nc_vlen_t))
+                for i in range(totelem):
+                    vldata[i].len = 0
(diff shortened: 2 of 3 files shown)


jswhit:
Could @dopplershift 's test from issue #527 be added to tst_vlen.py along with a Changelog entry?

ckhroulev:
@jswhit Absolutely!

ckhroulev:
Note that Travis CI uses NetCDF 4.1.1 while AppVeyor has 4.3.3.1.

I'll post an update once I have more info. (I'll have to install NetCDF and HDF5 versions matching ones used by Travis CI scripts.)

jswhit:
4.3 can be installed in travis ci trusty

https://launchpad.net/~heiko-klein/+archive/ubuntu/netcdf

jswhit:
The travis tests still fail even with hdf5-1.8.15-patch1 and necdf-4.4.0 (see pull request #606).

Works fine for me on OSX.  Maybe it's related to the compiler toolchain?

jswhit:
I'd like to go ahead and merge this, since it is the right thing to do. Let's just disable the test for right now until we figure out why it fails on Ubuntu.

ckhroulev:
@jswhit If I understand this correctly, the test I added will fail with NetCDF versions < 4.4.1 (see https://github.com/Unidata/netcdf-c/issues/221). I have no idea why it passed on AppVeyor (I blame it on the [undefined behavior of `memcpy` when one of the pointer arguments is null](http://en.cppreference.com/w/c/string/byte/memcpy)).

jswhit:
Let's wrap the test with `if __netcdf4libversion__ > '4.4.0'` then.

ckhroulev:
@jswhit Regarding turning the test off for NetCDF < 4.4.1: I did this already ([please see the code](https://github.com/ckhroulev/netcdf4-python/blob/issue527/test/tst_vlen.py#L140)).

(One of AppVeyor checks failed during "`git clone`". The rest passed, skipping this test, as expected.)

jswhit:
OK, great.  Merging now then.

PR #606, opened 2016-11-17
switch to trusty to get netcdf 4.3

jswhit:


ckhroulev:
I think it is helpful to have an automatic test (in Travis CI or similar) that uses the oldest supported NetCDF version (currently 4.1.1, if I'm not mistaken).

What's your take on this?

jswhit:
Yes, I agree in principle, but I also think 4.1.1 is too old to bother with.

ckhroulev:
Agreed.

What should the "oldest supported version" be then? Should we replace "4.1.1" with "4.x.y" in documentation? (After you pick x and y, that is.)

Also, @shoyer recently brought up dropping Python 2.6 support. What do you think about that?

jswhit:
I agree with dropping python 2.6. I see that  netcdf-c 4.1.3 is the default version in trusty.

jswhit:
I'd like to go ahead and commit this, since it is the right thing to do.  Let's just disable the test for right now until we figure out why it fails on Ubuntu.

PR #629, opened 2017-01-08
Update AppVeyor configuration.

### LINKED ISSUE #628: appveyor builds failing
on master:

```
running run_test.py
OC xxdr depends on the assumption that sizeof(off_t) == sizeof(void*)
  File "C:\Miniconda\conda-bld\conda.recipe_1483884027499\test_tmp\run_test.py", line 46, in <module>
    os.chdir(test_dir)
WindowsError: [Error 2] The system cannot find the file specified: 'C:\\Miniconda\\conda-bld\\conda.recipe_1483884027499\\work\\test'
```

@ocefpaf, any idea what's happening here?

ocefpaf:
Closes #628

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,37 +1,23 @@
 environment:
 
-  CONDA_INSTALL_LOCN: "C:\\conda"
-
   # SDK v7.0 MSVC Express 2008's SetEnv.cmd script will fail if the
   # /E:ON and /V:ON options are not enabled in the batch script interpreter
   # See: http://stackoverflow.com/a/13751649/163740
   CMD_IN_ENV: "cmd /E:ON /V:ON /C obvci_appveyor_python_build_env.cmd"
 
-  # Workaround for https://github.com/conda/conda-build/issues/636
-  PYTHONIOENCODING: "UTF-8"
-  PYTHONUNBUFFERED: "1"
-
-  # We set a default Python version for the miniconda that is to be installed.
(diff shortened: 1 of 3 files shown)


appveyor.yml
```diff
@@ -1,37 +1,23 @@
...
-  PYTHONUNBUFFERED: "1"
-
-  # We set a default Python version for the miniconda that is to be installed.
-  # This can be overridden in the matrix definition where appropriate.
-  CONDA_PY: "27"
-
   matrix:
-    - TARGET_ARCH: x86
-      CONDA_NPY: 111
-      CONDA_PY: 27
```
ocefpaf: I updated the configuration and remove the `x86` tests and Python 3.5. I believe that `x64` and both Python 2.7 and 3.6 is enough but let me know if you disagree and I can add them back.

conda.recipe/meta.yaml
```diff
@@ -1,45 +1,47 @@
...
 test:
-    imports:
-        - netCDF4
-        - netcdftime
-    commands:
-        - ncinfo -h
-        - nc4tonc3 -h
-        - nc3tonc4 -h
+  source_files:
+    - test
```
ocefpaf: This was the cause of the failure. Latest `conda-build` no longer accepts `SRC_DIR` env var in test. We need to copy the test folder with the `source_files`.

jswhit:
Brilliant - thanks! I never would have figured that out...

Issue #637, opened 2017-02-22
Incorrect timestamp format in netCDF4 API Documentation


djgagne:
In the [netCDF4 API docs Section 7](http://unidata.github.io/netcdf4-python/#section7), the text currently reads "Time coordinate values pose a special challenge to netCDF users. Most metadata standards (such as CF) specify that time should be measure relative to a fixed date using a certain calendar, with units specified like hours since **YY:MM:DD hh-mm-ss**." The dashes and colons on the example timestamp should be reversed and should read "YY-MM-DD hh:mm:ss". The time formats in the example code below this paragraph look correct.

jswhit:
fixed in pull request #638 - thanks for catching this!

PR #638, opened 2017-02-23
fix for docstring bug (issue #637)

jswhit:


djgagne:
The separators between YYYY, MM, and DD should also be changed from : to -, so it should look like YYYY-MM-DD hh:mm:ss.

netCDF4/_netCDF4.pyx
```diff
@@ -1294,11 +1295,14 @@ cdef _set_att(grp, int varid, name, value,\
                 try:
                     if force_ncstring: raise UnicodeError
                     dats_ascii = _to_ascii(dats) # try to encode bytes as ascii string
+                    print 'write as text 2'
```
dopplershift: I'm pretty sure you didn't mean to commit these print statements either...

Issue #406, opened 2015-05-05
Reading Dataset from memory


m-novikov:
Can netcdf4 read Dataset from memory?
Something like this:

``` python
from netCDF4 import Dataset
f = open('netcdf_file.nc', 'rb')
df = Dataset(f)
```

It would be quite convenient and useful API.
As for now to read netCDF4 Dataset from URL I need explicitly save it as named temporary file.

lesserwhirls:
Greetings!

We are working on this on the C side at Unidata, and hopefully it will be
finished soon. Once it's in the C lib, there should be very little work to
expose it in the netcdf4-python lib.

Cheers!

Sean

dopplershift:
To be clear, though, this would involve passing the entire "string" of data, not a file-like object.

jswhit:
you can create a Dataset in memory, using 'diskless=True'.

lesserwhirls:
Create, yes, but read - no.

Let's say you use the NetcdfSubset service from the THREDDS Data Server.
The server will return a netCDF file. When you use urllib2 to make the
request, you end up reading the server response, which is the bytes of a
netcdf file already in memory. The idea is to read in memory to remove the
need to write a temporary file to disk.

Sean

jswhit:
you could copy the data directly to a diskless file (without first writing to disk) couldn't you?

dopplershift:
Define "diskless file".

jswhit:
from netCDF4 import Dataset
nc = Dataset(URL)
ncm = Dataset('inmemory.nc',diskless=True,mode='w')
.....logic to copy data from nc to ncm...

then you have a 'diskless' or in-memory version of the dataset at URL

dopplershift:
Is the `nc = Dataset(URL)` supposed to work for anything besides opendap? It didn't for me when I just tried to hit a THREDDS server using NCSS. What I'm picturing is:

```python
from netCDF4 import Dataset
form urllib.requests import urlopen
url = urlopen(URL)
ncdata = url.read()
nc = Dataset(ncdata, diskless=True, mode='r')
```

m-novikov:
Passing string seems like fine idea, in my case of usage this will be convenient enough (I read 10mb NetCDF files which is not problem to store in python memory). 
But reading whole data in memory for python will increase memory consumption and this string will need to be garbage collected which is not always happens fast in CPython. 
[...]

Issue #295, opened 2014-10-05
Passing file objects to netCDF4.Dataset doesn't work


rabernat:
I am trying to port some code from scipy.io.netcdf_file to netCDF4.Dataset. I have encountered an issue which is pretty significant for me. netCDF4.Dataset expects a string as its argument and is unable to accept an open file object. The issue can be seen in the following code

``` python
import netCDF4
from scipy.io import netcdf_file

fobj = open('MODIS.nc', 'rb')
nc3 = netcdf_file(fobj)
fobj.close()

fobj = open('MODIS.nc', 'rb')
nc4 = netCDF4.Dataset(fobj) # this fails
fobj.close()
```

The second-to-last line raises

```
TypeError: expected string or Unicode object, file found
```

This may seem like an unnecessary feature (why not just pass the filename directly), but the problem is that I have a large archive of bzipped netcdf files on disk. The way I usually read them is

``` python
import bz2
bz2_fobj = bz2.BZ2File('MODIS.nc.bz2')
nc3 = netcdf_file(bz2_fobj)
```

If I can't do this with netCDF4, I will have do design a clumsy workaround involving system commands to manually unzip the files.

I considered tying to add this feature myself, but then I realized that the whole library was written in C. Hopefully you will consider adding support for reading file objects.

niallrobinson:
Hi everyone - I'd love to see a fix for this. Its coming up quite regularly in "map reduce" world (i.e. Hadoop, Spark, Dask) where we want to be able to pass file objects around and read them quickly, that is without dumping to disk. Is there anything on the horizon that might help out with this?

dopplershift:
netCDF-C last fall gained support for reading directly from an in-memory buffer that contains the bytes of a netCDF file. It's been on my TODO list to expose this in the Cython wrappers here, but I haven't gotten to it. That's probably your best bet--it's not a file-like object, but at least you wouldn't have to have a file on disk any more.

niallrobinson:
great - thanks for the update

jgerardsimcock:
@dopplershift any update on this?

rabernat:
As the original creator of this issue, I am pleased to see it is still alive. I am still very interested, although more for the reasons described by @niallrobinson. I believe the in-memory buffer solution could solve things. To clarify, would be be able to pass a BytesIO object?

niallrobinson:
yup - still actively thinking/worrying about this ;)

dopplershift:
[...]

PR #652, opened 2017-04-28
add support for opening file from memory

thehesiod:
fixes: https://github.com/Unidata/netcdf4-python/issues/406 and https://github.com/Unidata/netcdf4-python/issues/295, requires: https://github.com/Unidata/netcdf-c/issues/394 to be fixed


thehesiod:
btw unittests are failing because it can't find the nc_open_mem routing, not sure if this should be dynamically looked up or bump up the minimum netCDF version required

jswhit:
This is great, thanks!

To fix the issue with the tests, take a look at the `check_api` function in `setup.py`.  You can add a check there to see if the include file has `nc_open_mem`.  setup.py should then modify `include/constants.pyx` and add `DEF HAS_NC_OPEN_MEM = 1`.  The`nc_open_mem` calls in `_netCDF4.pyx` can then be wrapped with `IF HAS_NC_OPEN_MEM:`.

thehesiod:
btw, why is netCDF4/_netCDF4.c checked in?  it's generated when setup.py is run.

jswhit:
We'll also need a test.

_netCD4.c has to be manually generated.  It should be created with all the values in include/constants.pyx set to 0.  It's only used if cython is not installed.

jswhit:
Probably we should just require cython and get rid of _netCDF.c in the repo.

matthew-brett:
Cython is super-easy to install these days - there are wheels for many versions for OSX, Windows, Linux.

thehesiod:
working on updates, btw mind if I PEP'ify some of the files?  get a lot of warnings in PyCharm

thehesiod:
ok, made nc_open_mem lookup dynamic and added a test.

thehesiod:
btw, I'm not even sure how the .c file here helps because isn't it specific to a python version / platform / cython version?  Not sure how that ever worked correctly.

thehesiod:
btw let me know how I should fix this new unittest error.  I'm guessing it's compiling with all the settings off.  Further it won't work until there's a version of netcdf-c with the fix

shoyer:
> Probably we should just require cython and get rid of _netCDF.c in the repo.

The standard way to do this is to generate C files for upload to pypi, but not include them in the source.
cytoolz has a relatively sane version of this setup: https://github.com/pytoolz/cytoolz/blob/master/setup.py

netCDF4/_netCDF4.pyx
```diff
@@ -1712,7 +1721,7 @@ references to the parent Dataset or Group.
     the parent Dataset or Group.""" 
 
     def __init__(self, filename, mode='r', clobber=True, format='NETCDF4',
```
shoyer: Can we make `filename` optional if `memory` is provided?

Even if netCDF-C needs a bogus `filename` argument, I imagine we could generate something at random.

thehesiod: IMHO since `nc_open_mem` takes a filename we should use it

shoyer: OK, I guess so. It just seems pretty pointless to me :)

thehesiod: ya, weirdness :)

jswhit:
Looks like the `_netCDF.c` still uses` nc_open_mem`, so that will cause a failing test.  To generate one that will work with the minimum version of the netcdf library, edit `constants.pyx` and set all the elements to zero, then `cd _netCDF; cython -I../include _netCDF4.pyx`, then commit the new version of `_netCDF.c`.

I'd like to remove _netCDF.c from the repo, but that should be a separate pull request.

Are the other failing tests due to the bug in Unidata/netcdf-c#394?

thehesiod:
so I think the tests are failing now because it's running the mem test with the feature disabled?

jswhit:
The test fails on my machine even though I have 4.4.1 (with nc_open_mem).   I guess this is because of Unidata/netcdf-c#394?  I guess we should disable the test in run_all.py if ``__netcdf4libversion__`` < 4.4.1.2 (assuming that the fix gets in the next release of netcdf-c).

thehesiod:
btw, it seems like we should be able to use a memoryview insead reading http://docs.cython.org/en/latest/src/userguide/memoryviews.html?highlight=memory%20views.html.  Let me know what you guys think

netCDF4/_netCDF4.pyx
```diff
@@ -1776,17 +1784,27 @@ references to the parent Dataset or Group.
...
+        
+        **`memory`**: if not `None`, open file with contents taken from this block of memory.
+        Must be a sequence of bytes.  Note this only works with "r" mode.
         """
         cdef int grpid, ierr, numgrps, numdims, numvars
         cdef char *path
         cdef char namstring[NC_MAX_NAME+1]
+
+        memset(&self._buffer, 0, sizeof(self._buffer))
+
```
jswhit: Probably ought to add `_buffer` to `_private_atts` so it doesn't get mistaken for a netcdf attribute.

jswhit:
Don't see any real benefit to using memoryviews, since we're not slicing or manipulating the buffer in any way.

netCDF4/_netCDF4.pyx
```diff
@@ -1054,9 +1055,11 @@ default_fillvals = {#'S1':NC_FILL_CHAR,
 is_native_little = numpy.dtype('<f4').byteorder == '='
 is_native_big = numpy.dtype('>f4').byteorder == '='
 
-# hard code this here, instead of importing from netcdf.h
+# hard code these here, instead of importing from netcdf.h
 # so it will compile with versions <= 4.2.
 NC_DISKLESS = 0x0008
+NC_INMEMORY = 0x8000
```
jswhit: Doesn't look like `NC_INMEMORY` is ever used

netCDF4/_netCDF4.pyx
jswhit: Some text should be added to the docstring noting that `filename` is only used to set the value of `filepath` when `memory` is specified.

test/tst_open_mem.py
```diff
@@ -0,0 +1,23 @@
...
+import unittest
+import netCDF4
+
+CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
+
+
+class TestOpenMem(unittest.TestCase):
+    def test_mem_open(self):
+        # Needs: https://github.com/Unidata/netcdf-c/pull/400
+        if netCDF4.__netcdf4libversion__ >= '4.4.1.2':
```
jswhit: Why not just add this check in run_all.py so the test is skipped altogether?

netCDF4/_netCDF4.pyx
thehesiod: removed

test/tst_open_mem.py
thehesiod: actually updated the test to test different scenarios of this

netCDF4/_netCDF4.pyx
thehesiod: added

jswhit:
Merging now - thanks again for this contribution!

thehesiod:
thank you for the quick responses!  pleasure!

TWellman:
Is there a working example/documentation based on this discussion for pulling a dataset via a url into memory using the newest netCDF4 version? I've read material on adding a memory=bytes construct in the __init__ function to use the in-memory read capability from nc_open_mem(). However, my initial attempts have produced file location errors - "No such file or directory". I also ran a portion of the test script provided at https://github.com/Unidata/netcdf4-python/blob/master/test/tst_open_mem.py. An error was raised for netCDF4.__netcdf4libversion__ = '4.4.1.1' (< 4.4.1.2). Do I need to pull from https://github.com/Unidata/netcdf-c/pull/400 and integrate netcdf library changes directly into the package? Clarification is appreciated!

TWellman:
I am testing on a Jupyter notebook using Python 3.5 and netCDF4 version 1.3.0, but could switch if needed.

jswhit:
I think you need to link against the updated netcdf-c library.  This one should work:

https://github.com/Unidata/netcdf-c/releases/tag/v4.5.0-rc1

dopplershift:
Another option is to pass it a filename to an empty file, like from `tempfile.NamedTemporaryFile`

TWellman:
Great, substituting a completely unrelated local netCDF file as the "filename" but also passing a memory file in byte format proved to be successful. The returned file dimensions were consistent with the memory file rather than the unrelated local netCDF file with different dimensions/variables. So that checks out.

I also attempted **unsuccessfully** to use the temp file option using FILE_NAME = tempfile.NamedTemporaryFile(suffix='.nc', delete=False).name, mainly because it is much cleaner in my view. However, using a temp file as prescribed returned an OSError: NetCDF: Unknown file format. Should I format the temp file in a specific way? Thank you.

thehesiod:
I think "filename" just needs to be non-empty string

TWellman:
If I just input a text string the code returns an OSError: No such file or directory, e.g. using nc_file = netCDF4.Dataset('temp_file', mode = 'r', memory=bytefile)

thehesiod:
what version of netcdf-c ?

dopplershift:
@TWellman I needed it to be a netCDF file (apparently--I guess I forgot about it). This is what I've done for now until netCDF 4.5 is released:
```python
    with NamedTemporaryFile(suffix='.nc') as temp:
        # Create a temporary netCDF file to work around bug in netCDF C 4.4.1.1
        # We shouldn't actually need any file on disk.
        nc_temp = Dataset(temp.name, 'w').close()
        nc =  Dataset(temp.name, memory=mem_buffer)
```

TWellman:
I should probably upgrade as suggested earlier. I am using netCDF4.netcdf4libversion = '4.4.1.1', netCDF4 version 1.3.0.

TWellman:
Great, thank you!

Issue #654, opened 2017-04-30
Dataset class should support encoding parameter to override global attribute


thehesiod:
In fact the whole idea of having a global `encoding` property is a bad idea because then you can't support multiple Datasets with different encodings.  The real bug should be to deprecate `netCDF4.encoding`

as an example, the MADIS meso files are encoded in what appears to be cp1252.

thehesiod:
if someone from the team can give feedback on how they'd like this achieved I can work on this.

jswhit:
I suggest adding a `set_encoding` method to override the global value. There are already too many kwargs in `Dataset.__init__`.

thehesiod:
What if you need the encoding before `set_encoding` is called? Could that ever be possible? It really does feel like an init param. Will look into `unicode_error`

thehesiod:
ok added support for `encoding_errors` looking into why unittests are failing

shoyer:
Does netCDF really support arbitrary encodings for strings but not have any way of indicating them in the data model? That seems like a disaster waiting to happen...

thehesiod:
from what I understand yes!  I ran into this with MADIS mesonet dataset that had the "Annœullin" string in it (has character \x9c in it), and from what I saw there was nothing specifying the encoding :(  After some grepping around it seemed like this was most likely a CP1252 format from someone generating the files on a windows box.   some info: http://www.unidata.ucar.edu/software/netcdf/docs/BestPractices.html#bp_Strings-and-Variables-of-type-char   basically the format was not designed with a true "string" type, only added in netCDF 4.x.

jswhit:
Even with the netcdf-4 `NC_STRING` type there is no concept of an encoding in the data model. It's just stored as a string of bytes in the file, and the client has to know how to encode it into a string.  I can't find anything in the CF metadata standard that relates to string encoding, so there's no standard way for the client to figure out how to encode it.  The python interface always returns a string, and the encoding is currently defined by a global module variable.  I gather the purpose of pull request #655 is to at least allow the user to change the encoding on a per-Dataset basis, so multiple Datasets can be accessed at once with different encodings specified.   

When it comes to  names of variables, dimensions, attributes, groups, and types netcdf-c always uses UTF-8 encoding.

jswhit:
Just noticed this at http://www.unidata.ucar.edu/software/netcdf/docs/file_format_specifications.html

[...]

PR #655, opened 2017-04-30
Add per Dataset encoding support

thehesiod:
- resolves https://github.com/Unidata/netcdf4-python/issues/654
- fixes warning related to `nc_open_mem` change
- bumps up cython to 0.20 for `basestring` support


thehesiod:
btw to make it more backwards compatible we could instead change the default encoding to `None`, with a dynamic fallback in the `encoding.__get__`  lookup to `default_encoding` instead of fixing it during `__init__`.  thoughts?

thehesiod:
hah, that recursion bug was a doozy, put a warning for future devs

thehesiod:
cool, looks like this PR is ready for review!

jswhit:
I'm waiting on this to see how the discussion on netcdf-c plays out (https://github.com/Unidata/netcdf-c/issues/402)

jswhit:
What if different NC_STRING variables within the same dataset contain data with different encodings?  I suppose this could be handled with the proposed `_Encoding` attribute. 

The whole thing is a mess and any solution I can think of (including this one) seems fragile and kludgy.  At least we have a solution now that works as long as you know the encoding and you only are reading data from one Dataset at a time.

thehesiod:
ya, we need per variable encoding fallbacks.  Let me know how you'd like to handle that.  Some ideas:
- user specified callback that takes one parameter (variable name) and returns encoding to use in case `_Encoding` not specified and it's a char type.
- a `Dataset.get_variable` method which you can use to specify and encoding

jswhit:
Is your use case (for specifying an encoding) mainly for attributes, or variable data?

thehesiod:
variable data, based on the conversations in the netcdf-c thread it sounds like attribute should all be forced to utf-8 (which I can change in this PR too).

jswhit:
The reason I ask is if you are mainly concerned about attributes, we could add a kwarg `encoding` to getncattr.  You would then have to use getncattr to retrieve string attributes, but you could specify the encoding on a per attribute basis (no need for a Dataset encoding parameter).

jswhit:
Could you post one of the MADIS files that you are dealing with?

thehesiod:
https://madis-data.cprk.ncep.noaa.gov/madisPublic1/data/archive/2017/02/01/LDAD/mesonet/netCDF/20170201_0000.gz

jswhit:
thanks @thehesiod.  I've created an alternate 'solution' in the 'encoding' branch.  Instead of setting the encoding as a Dataset init parameter, I look for an _Encoding attribute for character and vlen string variables.  For string variables, if _Encoding is not set, 'utf-8' is used.  For character arrays, if _Encoding is set, then a numpy array of fixed length strings is returned by automatically calling chartostring (the rightmost dimension of the variable is assumed to the the length of the strings).  If _Encoding is not set, you get the previous behavior (an array of single characters is returned).   So, in your case you would have to add `_Encoding="cp1252"` as a variable attribute, either using NCO or by opening the file in append mode and adding the attribute before you read the data.

For attributes, I added a 'encoding' kwarg to getncattr.

I know adding a new attribute to all the files is probably not a good solution for you, but I'm still a bit confused about what the problem is you are trying to solve.  With the current master, you can read the variable stationName and convert it to an array of strings using chartostring - but it will use the global module variable 'default_encoding'.  Wouldn't simply adding a kwarg 'encoding' to chartostring solve your problem?  (this is also done in the 'encoding' branch).

To be specific, here's what I'm suggesting:

```python
from netCDF4 import Dataset, chartostring
nc = Dataset('20170201_0000')
chararr = nc['stationName'][:]
strarr = chartostring(chararr,encoding='cp1252')
print strarr[77]
nc.close()
```

thehesiod:
ya thinking about it more doesn't make sense to need an encoding init param for chat variable data, not sure why I thought I needed this.  Closing this PR

Issue #656, opened 2017-05-04
Interpretation of reserved _Unsigned attribute written by netCDF-Java


deeplycloudy:
Unsigned integer data written by NetCDF-Java does not use an unsigned integer NetCDF type (e.g. `uint16`), but rather sets the reserved attribute `_Unsigned` on a (signed) `int16` type. The result for the user of netcdf4-python is data which must be manually corrected after being read in.

It is my understanding that netCDF-C tries to "do no magic" so this is not a netcdf4-python bug per-se.  #493 also notes the preference of netCDF4-python for leaving metadata interpretation to downstream applications. However, this is not a metadata convention in the CF sense, but rather a documented low level implementation decision within netCDF-Java: http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/CDM/Netcdf4.html
> **Differences between netCDF-4 C and Java libraries for netCDF4 files**
>*Unsigned types*
>The C library uses unsigned integer types: NC_UBYTE, NC_USHORT, NC_UINT, NC_UINT64.
The Java library does not have separate types for unsigned integers, but adds the reserved attribute _Unsigned = "true" when the variable is unsigned. One can check this with Variable.isUnsigned(), Attribute.isUnsigned(), and Array.isUnsigned(). Conversions done by the library are aware of this convention. Java does not have unsigned types, and we dont want to double the internal memory requirements by widening the data.

I am in favor of a feature to assist the users in correcting for signendess on read when this attribute is present. One possible route:
- Add a method `set_auto_signededness(True)` to request the conversion be performed.
- If the above is set, perform the conversion if requested within the read slice method. 

I would be less in favor of adding a method to handle the conversion; it means every time a user reads data they must remember to call that method. But that would be more helpful than having every user write their own when encountering unsigned data written by netCDF-Java.

Thanks to @lesserwhirls for helping me understand some of the things above; any errors remain mine!

jswhit:
What exactly does the java library do when it reads data with `_Unsigned=True`?  Does the C library recognize the this special attribute? - I don't see any mention of it in the NUG.

jswhit:
I suppose we could return a view of the array with the appropriate unsigned dtype if the variable has `_Unsigned=True`.  I don't like the idea of adding a `set_auto_signededness` - if the attribute is set there should be no ambiguity about what is expected.  

On the other hand, it's not very hard for a user to just do `data = data.view(np.uint32)`...

deeplycloudy:
[...]

PR #658, opened 2017-05-08
recognize _Unsigned attribute and return unsigned integer data (issue #656)

jswhit:


netCDF4/_netCDF4.pyx
```diff
@@ -3822,6 +3822,13 @@ rename a `netCDF4.Variable` attribute named `oldname` to `newname`."""
             # else if variable has only add_offset attributes, rescale.
             elif hasattr(self, 'add_offset') and self.add_offset != 0.0:
                 data = data + self.add_offset
+
+        # if attribute _Unsigned is True, and variable has signed integer
+        # dtype, return view with corresponding unsigned dtype (issue #656)
+        is_unsigned = getattr(self, '_Unsigned', False)
+        if is_unsigned and data.dtype.str[1] == 'i':
```
shoyer: `data.dtype.kind == 'i'` is more robust here

jswhit: right - done.

jswhit: I don't think the `Variable.dtype` should be updated - since it should reflect the data type of the netcdf data stored in the file.  `Variable.dtype` plus `_Unsigned=True` imply that the returned data will be unsigned.  I agree that this might be problematic since the returned data now has a type different than what is in the file.  We already do this with data packed as short integers with `scale_factor` and `add_offset` though.

shoyer: It is an important invariant (relied on by downstream tools like xarray) that `variable[...].dtype == variable.dtype`, at least if `set_automaskscale(False)`.

Possibly that flag (or another one?) should also control where this sort of behavior is done? Xarray prefers a low-level interface, with minimal magic.

deeplycloudy:
I think there may be an odd interaction of this fix with add_offset and scale_factor conversion. The first case below demonstrates it's doing the unsigned conversion correctly, but that doesn't seem to work well in full-auto mode.

This gives me the right answer of -139.505, -43.7424:
<pre>
event_lons = nc.variables['event_lon']
event_lons.set_auto_scale(False)

scale_factor = event_lons.scale_factor 
add_offset = event_lons.add_offset
event_lons_fixed = (event_lons[:])*scale_factor+add_offset

print(event_lons_fixed.min())
print(event_lons_fixed.max())
</pre>


While this gives the wrong answer of -222.566, -89.5478:
<pre>
event_lons = nc.variables['event_lon']
event_lons_fixed = event_lons[:]
print(event_lons_fixed.min())
print(event_lons_fixed.max())
</pre>

deeplycloudy:
The fix for the add_offset and scale_factor problem above is to move the unsigned conversion to immediately before the block where the scale factor is applied. This is the correct place to perform the conversion because signed integers lose meaning once scaling is applied.

deeplycloudy:
To be explicit: copy-paste the `_Unsigned` block to line #3795 of _netCDF4.pyx.

jswhit:
OK - done.

deeplycloudy:
I tried the revised PR, and confirm it fixes the issue I previously raised.

shoyer:
We still need a way to disable this behavior (preferably reusing `set_automaskandscale`). If this behavior is triggered, it will break downstream libraries like xarray in subtle and hard to understand ways.

jswhit:
How about adding `set_auto_unsigned` with a default of `False`?

shoyer:
> How about adding set_auto_unsigned with a default of False?

That would work for me, too, though I suspect that in practice it would make this change pointless. When you know to look for `_Unsigned`, the fix in user code is quite easy.

I like reusing `set_automaskandscale` because it's already used for selecting between low-level (raw data) and high-level (data fixed up based on attributes) APIs. I don't see many use cases for a mix of low-level and high-level behavior, and in a way, `_Unsigned` is a form of automatic scaling, so this shouldn't be too surprising.

This way we can make the default `True`, which means that users are much more likely to actually *use* this option. Libraries like xarray that the raw data probably already turn off automatic masking and scaling, so this won't break them.

jswhit:
OK, I added the check for autoscaling.  This behavior can now be turned off with `set_auto_scale(False)` or `set_auto_maskandscale(False)`.

jswhit:
@shoyer, are you OK with this now? If so, I'll go ahead and merge.

shoyer:
Yes, looks good to me!

PR #661, opened 2017-05-12
Fix misspelled IndexError in utils

scriptator:
This fix addresses the following NameError: 'IndexErro' is not defined


jswhit:
Good catch - thanks!

PR #666, opened 2017-05-19
add dependency on cython >= 0.19.

jswhit:
remove _netCDF4.c, _netcdftime.c and constants.pyx from repo


davidh-ssec:
Is there a reason this was done this way? I now getting failing jobs on travis when trying to pip install netcdf4. See https://travis-ci.org/pytroll/satpy/jobs/239619486

Note the no binary flag was due to a buggy wheel (I think), but it's been a while so I could be wrong.

When installing without cython already installed this results in a importerror instead of pip knowing about the dependency and downloading/installing it.

jswhit:
It was done to get rid of the need to have the c source files in the repo.   I think pull request #673 should fix this.

davidh-ssec:
Why wouldn't you want the C files in the repository? Unless things have changed this is how most projects that use cython are set up. You try to import cythonize, if you can't then compile the C/C++ source files that are already available. Or is this an issue for NetCDF4 since you may need to change the cython depending on the version of the C library installed?

jswhit:
Because the C source files in the repo were the lowest common denominator, assuming you had a very old version of the netcdf lib.  Now setup.py can determine what features are available, and then enable them when the C source file is created using cythonize.

It seems like cython is pretty easy to install, and indeed already installed on most systems  - so why not require it?  Do you see this as a problem?

davidh-ssec:
> so why not require it?

My only real complaint is that it is another dependency and that it is not the usual case to require users to install it. I can see that if you are dynamically determining available C library features at installation time it makes sense to compile things based on those features. However, does this mean that if a user upgrades or changes their version of the C library that `netcdf4-python` won't be compatible or won't provide all the possible features?

My only other issue is that as it is currently in PyPI a user can't install NetCDF4 in one go unless they already have Cython installed (otherwise they have to get an ImportError, install Cython, and then install netcdf4-python again0.

jswhit:
Yes, if you upgrade the netcdf lib you may have to rebuild netcdf4-python to get any new features.  But if we provide the C source, it won't have any of those features either.

We had been providing the C code and using it if cython was not installed for many years.  Many users complained that they weren't getting the new features, and developers were forgetting to re-generate the C source code in the repo with all new features disabled.   I could be convinced to go back to the old approach if requiring cython really was a hardship - but it seems pretty minimal to me in comparison to the HDF5 and netCDF C lib dependencies.

I don't know much about pip, but I wonder if there is any way to tell it about the cython dependency so that it will install it first, before setup.py tries to import cython?

thehesiod:
check this out: https://github.com/h5py/h5py/issues/535 apparently newer versions of pip may resolve this.

davidh-ssec:
@jswhit That makes sense. I guess you would either have to require a certain minimum version of the C library and make sure to update the generated C code. I could also see it working to go against the normal cython pattern and if it is present (either installed or installable) then regenerate the C code. I agree with you that it is not a huge installation problem to require cython given how it is being used.

However, it would be cool if the cython code (and therefore the generated C code) could handle a wide range of C versions but I know I wouldn't want to try to maintain that so I don't expect you guys to. I am also not familiar with all the ways you are investigating the C library to prepare the cython code so I don't think I'll be of much help on updating the setup.py file until I see what's going on currently.

davidh-ssec:
@thehesiod @jswhit Looks like h5py has a good solution for dynamically/building the cython extensions on installation without importing cython in the setup.py directly: https://github.com/h5py/h5py/blob/master/setup.py#L163

jswhit:
I've updated setup.py in pull request #673 to use the trick suggested in https://stackoverflow.com/questions/37471313/setup-requires-with-cython, which is to not import cython directly in setup.py, but specify it in setup_requires along with setuptools >= 18.0 (which can handle .pyx files directly).

@david-ssec, would you mind trying this to see if it fixes the `pip install` issue?

davidh-ssec:
@jswhit I've changed my package's travis config to install netcdf4-python from the issue666 branch. It passes on python 3.5 and 3.6. It fails but on 2.7, 3.3, and 3.4 due to a setuptools incompatibility. See https://travis-ci.org/pytroll/satpy/jobs/239944247

I then forced the update of setuptools before installing everything else and that passed on all versions: https://travis-ci.org/pytroll/satpy/builds/239948942

Looks like it is good to go.

jswhit:
Great - I'll merge this later today.

PR #673, opened 2017-06-05
fix for issue #666 (pip fails to install cython dependency)

jswhit:


davidh-ssec:
Isn't this still importing Cython though? If it isn't installed it's still going to raise an ImportError.

PR #674, opened 2017-06-06
preparing for v1.2.9 release

jswhit:


.travis.yml
```diff
@@ -18,7 +18,7 @@ python:
   - "2.7"
   - "3.5"
   - "3.6"
-  - "3.7-dev"
+#  - "3.7-dev"
```
dopplershift: Instead of removing it, what about putting this under the matrix section below (same depth as `include`):
```yml
matrix:
  allow_failures:
    - python: "3.7-dev"
```

jswhit:
good idea - done.

davidh-ssec:
@jswhit Any idea when the expected merge and final release of 1.2.9 will be?

jswhit:
I was waiting a bit to see if a separate netcdftime release might happen, so that we could remove the bundled netcdftime in 1.2.9 - but I guess it's going to take a while longer.  I can create a release early next week.

Issue #677, opened 2017-06-15
HDF5 include path detection issue


thehesiod:
by default setup.py will use nc-config to detect all the include paths to use when cythonizing the pyx file.  This is not completely correct because nc-config will only include headers for linking against libnetCDF, and not HDF5.  Given that HDF5 is included directly (H5public.h), the include for this header should always be searched for separately.  On debian for example this is in: /usr/include/hdf5/serial.

jswhit:
We only need hdf5 headers to get the version number - maybe a better solution is to figure out how to get the version number from a library call. nc-config does detect the hdf5 library path.

jswhit:
looks like we could use this H5get_libversion

https://support.hdfgroup.org/HDF5/doc/RM/RM_H5.html#Library-Version

and get rid of the H5public.h include in netCDF4.pxi.

jswhit:
Never mind - we still would need to access the hdf5 headers when compiling if we got the version number with a library call.  However, looking at the nc-config utility, it seems that the --cflags option returns an include path that includes both the netcdf-4 install path and the CPPFLAGS used when building netcdf-c (which should include the path to the HDF5 headers).  Too bad this doesn't work on Debian.

thehesiod:
I think the issue is that when building netcdf I built specifying the path to the HDF5 header, and not a cflag specifying a new include path.  If it works for you we can just close this...perhaps worth a comment that you need to build netcdf w/ HDF5 a certain way.

thehesiod:
I was building it this way, and it resulted w/o settings in nc-config:
```
cmake .. -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} -DHAVE_HDF5_H=/usr/include/hdf5/serial/hdf5.h -DMFHDF_H_INCLUDE_DIR=/usr/include/hdf -DENABLE_HDF4=ON
```

jswhit:
Yep - nc-config only adds CPPFLAGS when you build with ./configure;make;make install, not when you use cmake.  Sigh.

The problem with using the header to get the library version is that it is easy to pick up a header that is different from the one used to actually compile the library.  

I've created a new pull request (pull request #679) that adds your header search, but uses the H5get_libversion call to get the version number to avoid this problem.

thehesiod:
[...]

PR #678, opened 2017-06-15
fix HDF5 include directory search

thehesiod:
fixes https://github.com/Unidata/netcdf4-python/issues/677
also PEP'ifys the file


thehesiod:
looks like the failures are due to python 3.7 stuff unrelated to my changes...not sure about appveyor

PR #681, opened 2017-06-23
Unnecessary to use a slice when the stride length is larger than 1

fredrik-1:
The reading with a slice with a stride length larger than 1 is sometimes extremely slow. It is thus better to not change a sequence of indices to a slice in that case.

### DIFF
--- a/netCDF4/utils.py
+++ b/netCDF4/utils.py
@@ -260,7 +260,7 @@ Boolean array must have the same shape as the data along this dimension."""
             # The ellipsis stands for the missing dimensions.
             newElem.extend((slice(None, None, None),) * (nDims - len(elem) + 1))
             hasEllipsis = True
-        # Replace sequence of indices with slice object if possible.
+        # Replace sequence of indices with slice object if possible with a stride of 1
         elif np.iterable(e) and len(e) > 1:
             start = e[0]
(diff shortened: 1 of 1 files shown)


netCDF4/utils.py
```diff
@@ -269,7 +269,7 @@ def _StartCountStride(elem, shape, dimensions=None, grp=None, datashape=None,\
                 ee = range(start,stop,step)
             except ValueError: # start, stop or step is not valid for a range
                 ee = False
-            if ee and len(e) == len(ee) and (e == np.arange(start,stop,step)).all():
+            if ee and e[1]-e[0]==1 and len(e) == len(ee) and (e == np.arange(start,stop,step)).all():
```
jswhit: code would be clearer if with `step` instead of `e[1]-e[0]`

jswhit: How about 

            if ee and len(e) == len(ee) and (e == np.arange(start,stop,step)).all():
                if step != 1 and len(e) < 1000: # don't convert to slice
                    newElem.append(e)
                else:  # convert to slice
                    newElem.append(slice(start,stop,step))

jswhit: a step of -1 should be convert to a slice also, so the check should be `step not in [1,-1]`

Issue #680, opened 2017-06-21
Very slow when using indexing with a constant step size, for example [0, 2, 4]


fredrik-1:
Reading a netcdf variable with for example the index 0:10:2 is several hundreds times slower than with a index with a non constant step size. 

nf=Dataset('file.nc', 'r')
v=nf['temperature'][0:10:2, :, :] takes much longer time than
v=nf['temperature'][0:10, :, :]

The same is also true when using [0, 2, 4], [0, 3], or np.array([0, 5, 10]). It is also much faster to use
np.array([0, 5, 10, 10]) compared to using np.array([0, 5, 10])

I guess the problem is related to https://github.com/Unidata/netcdf4-python/issues/325 but I am not sure if it is supposed to be fixed or not.  

I have been using netcdf 1.2.4 and 1.2.9, the behavior is the same with both. 
I am using windows, anaconda download, ipython and python and conda install of netCDF4

jswhit:
If this slice is small, the overhead of the python interface (creating the start,count,stride arrays to pass to the C library) will be large relative to the cost of actually reading the data from the file.  I suspect that this is what is happening here.  You say it is several hundred times slower, but what are the actual timings?

jswhit:
Could you post your actual benchmark code (including the data file)?

fredrik-1:
The actual timing can be 10 seconds vs half an hour. The difference in time is also largest when the total download is relatively large (but the problematic slice might be small).

Here is the code with the first netcdf file found here 
https://www.unidata.ucar.edu/software/netcdf/examples/files.html

The reading is relatively small so the time difference is only about 4 times on my computer with that netcdf file. I am going to try to find a better example netcdf file.

from netCDF4 import Dataset
import time
import numpy as np

fileIn='ocean//sresa1b_ncar_ccsm3-example.nc'
nf=Dataset(fileIn,'r')
varN='tas' 
t0=time.time() 
q=nf[varN][:, 0:128, :]
print q.shape
t1=time.time()-t0
print 'it took: '+str(t1)

t0=time.time()
q=nf[varN][:, 0:200:2, :]
print q.shape
t2=time.time()-t0
print 'it took: '+str(t2)

print t2/t1

jswhit:
The reason the netcdf4 results are so slow has to be because of chunking:

http://www.unidata.ucar.edu/blogs/developer/entry/chunking_data_why_it_matters

fredrik-1:
[...]

PR #683, opened 2017-06-26
extend fix for issue #680 to include strided slices

jswhit:
Now strided slices are converted to individual calls to nc_get_vara, instead of one call to nc_get_vars. (as long as the total number of calls is less than 1000).


jswhit:
@shoyer - would you mind running the xarray tests on this before I merge?

shoyer:
I just started a build at https://travis-ci.org/pydata/xarray/builds/247647274, should have results in 10-15 minutes.

Look for the job "py36-netcdf4-dev"

shoyer:
The xarray test suite passes, which is encouraging but by no means definitive :)

PR #687, opened 2017-07-02
fixes #686 Encode filename with sys.getfilesystemencoding()      master (#1) 

### LINKED ISSUE #686: filename is encoded to utf8, resulting in broken filenames on windows
When i try to create a netcdf-file with a filepath, that contains Non-ASCII-characters, Netcdf4 creates broken filenames on windows, because it encodes the filepath in UTF8, which is fine for Linux,. In Windows, the filepath has to be convertet to CP1252.

```python
import netCDF4 as nc4
filepath = r'C:\Plön.nc'
ds = nc4.Dataset(filepath, 'w')
ds.close()
```

this creates a file 
PlÃ¶n.nc

I could try to avoid Non-ASCII-Characters in .nc filenames, but if i want to create a file in the folder
C:\Plön\
```python
import netCDF4 as nc4
filepath = r'C:\Plön\Ploen.nc'
ds = nc4.Dataset(filepath, 'w')
ds.close()
```
then i get the confusing error message:
builtins.OSError: Permission denied
Because the folder "Plön" is not recognized correctly.

The solution would be to pass the filesystemencoding to _strencode in 
_netCDF4.pyx in line 1796
```diff
-  bytestr = _strencode(str(filename))
+ bytestr = _strencode(str(filename), encoding=sys.getfilesystemencoding())
```
I'll create a pull request.
Max

MaxBo:


jswhit:
Looks reasonable.  Please add a Changelog entry and a test.

MaxBo:
I added a test. Where to i have to add the changelog entry?

jswhit:
In the Changelog file.

MaxBo:
added to Changelog under "version 1.3.0 (not yet released)".

jswhit:
Great - thanks!  Merging now..

jswhit:
tst_filepath.py is failing on my mac:

    FAILED (errors=1)
    [jeff-whitakers-imac-9:~/python/netcdf4-python.git/test] jsw% python tst_filepath.py
      File "tst_filepath.py", line 17
    SyntaxError: Non-ASCII character '\xc3' in file tst_filepath.py on line 17, but no encoding declared; 
    see http://python.org/dev/peps/pep-0263/ for details.

MaxBo:
as sourcefile encoding is missing for the test.
add
# -*- coding: utf-8 -*-

In addition, in line 1953 in _netCDF4.pyx
the return path should be decoded by the 
filesystemencoding, too

-  return py_path.decode('ascii')
+ return py_path.decode(sys.getfilesystemencoding())

filepath() could refer to an opendap URL. Can an opendap URL contain Non-ASCII-Characters and what encoding would it have? How to distiguish a filepath and an URL? 
Or just use try and exept?
try:
    return py_path.decode(sys.getfilesystemencoding())
except UnicodeDecodeError:
    return py_path.decode('utf-8')

i could try to fix this tomorrow.

jswhit:
OK, thanks.

jswhit:
opendap urls use standard 'percent encoding'

https://en.wikipedia.org/wiki/Percent-encoding

MaxBo:
I commited two changes in https://github.com/MaxBo/netcdf4-python.

The file excoding of the test and the decoding of the filepath in _netCDF4.pyx
```python
        if self.disk_format.startswith('DAP'):
            # OpenDAP-Datasource
            decoded_path = urllib.parse.unquote(py_path)
        else:
            # filepath on disk
            decoded_path = py_path.decode(sys.getfilesystemencoding())
        return decoded_path
```

I did not find an OpenDAP- test-dataset with an URL that includes a space or any other character, that would be "Percent-encoded". So i could not add a test for it.

Do i have to open a new Pull request, or is it automatically tested in this re-opend Pull-request?

jswhit:
Please open a new pull request

PR #691, opened 2017-07-06
fixes #686 Encode filename with sys.getfilesystemencoding()

### LINKED ISSUE #686: filename is encoded to utf8, resulting in broken filenames on windows

MaxBo:
fixes file encoding and decodes the filepath from the default filesystemencoding - or form percentage-encoding if it is a OpenDAV Datasource

### DIFF
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -1795,7 +1795,7 @@ references to the parent Dataset or Group.
         if diskless and __netcdf4libversion__ < '4.2.1':
             #diskless = False # don't raise error, instead silently ignore
             raise ValueError('diskless mode requires netcdf lib >= 4.2.1, you have %s' % __netcdf4libversion__)
-        if sys.platform == 'win32:
+        if sys.platform == 'win32':
             bytestr = _strencode(str_filename, encoding='mbcs')
         else:
             bytestr = _strencode(str_filename)
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -1957,7 +1957,7 @@ open/create the Dataset. Requires netcdf >= 4.1.2"""
                 else:
                     # filepath on disk
                     decoded_path = py_path.decode(sys.getfilesystemencoding())
-		return decoded_path
+                return decoded_path
         ELSE:
             msg = """
 filepath method not enabled.  To enable, install Cython, make sure you have
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -941,6 +941,7 @@ import netcdftime
 import numpy
 import weakref
 import sys
+import urllib
 import warnings
 from glob import glob
 from numpy import ma
@@ -1950,7 +1951,13 @@ open/create the Dataset. Requires netcdf >= 4.1.2"""
                 py_path = c_path[:pathlen] # makes a copy of pathlen bytes from c_string
             finally:
                 free(c_path)
-            return py_path.decode('ascii')
+                if self.disk_format.startswith('DAP'):
+                    # OpenDAP-Datasource
+                    decoded_path = urllib.parse.unquote(py_path)
+                else:
+                    # filepath on disk
+                    decoded_path = py_path.decode(sys.getfilesystemencoding())
+		return decoded_path
         ELSE:
             msg = """
 filepath method not enabled.  To enable, install Cython, make sure you have
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -1951,13 +1951,15 @@ open/create the Dataset. Requires netcdf >= 4.1.2"""
                 py_path = c_path[:pathlen] # makes a copy of pathlen bytes from c_string
             finally:
                 free(c_path)
-                if self.disk_format.startswith('DAP'):
-                    # OpenDAP-Datasource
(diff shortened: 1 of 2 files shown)


MaxBo:
Test test_filepath in appveyor failed, but checks have passed. Don't merge yet.
```
(C:\Miniconda-x64\conda-bld\netcdf4_1499351439744\_t_env) (root) C:\Miniconda-x64\conda-bld\netcdf4_1499351439744\test_tmp>"C:\Miniconda-x64\conda-bld\netcdf4_1499351439744\_t_env\python.exe" -s "C:\Miniconda-x64\conda-bld\netcdf4_1499351439744\test_tmp\run_test.py" 
..............F...............................................................................not running tst_unicode3.py ...
not running tst_dap.py ...
netcdf4-python version: 1.3.0
HDF5 lib version:       1.8.17
netcdf lib version:     4.4.1.1
numpy version           1.11.3
======================================================================
FAIL: test_filepath_with_non_ascii_characters (tst_filepath.test_filepath)
----------------------------------------------------------------------
  File "C:\Miniconda-x64\conda-bld\netcdf4_1499351439744\test_tmp\test\tst_filepath.py", line 24, in test_filepath_with_non_ascii_characters
    assert nc_non_ascii.filepath() == str(nc_non_ascii_file)
AssertionError
----------------------------------------------------------------------
Ran 94 tests in 33.469s
FAILED (failures=1)
```

jswhit:
typo in _netCDF4.pyx is causing compilation to fail.

MaxBo:
Sorry, Fixed. 
Using explicit distinction for the encoding between windows and linux/os now, because sys.getfilesystemencoding() returns 'utf-8' on the Appveyor-testsystem, which makes the test fail - maybe because it's a virtual machine.
Unittests pass now on Appveyor for python 2 and 3. 
On Travis, test_filepath is still skipped.

jswhit:
Can you explain what you are doing with this patch?  It appears that you assume that the filepath is always encoded as 'mbcs' on windows and 'utf8' on everything else.  

I found some discussion windows filename encoding in python here

https://www.python.org/dev/peps/pep-0529/#use-strict-mbcs-decoding

It was ultimately decided to use utf-8 and not mbcs.

jswhit:
Maybe a more general solution would be to provide an optional `encoding` kwarg to `Dataset.__init__` and `Dataset.filepath`, with the default set to `utf-8`.  Then we wouldn't have to make choices based on platform.

MaxBo:
Dear Jeff,

thanks for adding the kwargs-encoding parameter. This should work for me.

I assumed that windows filepaths are encoded as 'mbcs', because on my local windows machine, sys.getfilesystemencoding() returns mbcs. On the he AppVeyor-Test-Environment sys.getfilesystemencoding() returned utf8. 
In the test_filepath.py i used a filepath with a directoryname, that contained non-ascii-characters. 
```
tempdir = tempfile.mkdtemp(prefix=u'ÄÖÜß_')
```
When i encoded the filepath with utf8 (which sys.getfilesystemencoding() returned on Appveyor), the test failed with an OSError, because the tempdir was not found. When i used a hardcoded encoding='mbcs', the test passed. So i assumed that on windows the encoding was always mbcs, regardless what sys.getfilesystemencoding() returned.
I did not know the discussion in
https://www.python.org/dev/peps/pep-0529/#use-strict-mbcs-decoding

So your solution with the kwargs-parameter should be fine.
I'll add onother test with a directory containing Non-Ascii-characters.
Thanks,

Max

jswhit:
This is all very confusing.   Having an extra `encoding` kwarg argument seems the most flexible, and should in principle be able to deal with every situation.  However, it forces the user to know what the filename encoding is.  Your original solution sys.getfilesystemencoding() should work - it's just hard to create tests for systems with different default encodings that appveyor uses.  Also, it will fail for illegally encoded filenames (filenames will different encodings that the system default).  

So, I think we have three choices here:

1) use sys.getfilesystemencoding() to encode the filename in Dataset.__init__ and decode back into a string in Dataset.filepath.  No extra kwarg args added, keeps interface simpler - but may fail in some cases.

2) use an encoding specified as a kwarg argument in Dataset.__init__ and Dataset.filepath.  Forces user to know what the filename encoding is.

or 

3) as in (2), but use sys.getfilesystemencoding()  if no encoding kwarg is specified.

I've updated the pull request to use option (3).

MaxBo:
Thanks a lot - i think this should work now.

PR #702, opened 2017-08-21
Ellipsis bug, new version with error handling

fredrik-1:
The old fix resulted in the wrong error message if more than one ellipsis where used. The variable i was never used so I use the version I had written instead. I think it is slightly clearer but they are obviously very similar.

### DIFF
--- a/netCDF4/utils.py
+++ b/netCDF4/utils.py
@@ -177,18 +177,19 @@ def _StartCountStride(elem, shape, dimensions=None, grp=None, datashape=None,\
                 elem.append(slice(None,None,None))
     else:   # Convert single index to sequence
         elem = [elem]
-
+    
+    hasEllipsis=False
+    for e in elem:
+        if type(e)==type(Ellipsis):
+            if hasEllipsis:
+                raise IndexError("At most one ellipsis allowed in a slicing expression")
+            hasEllipsis=True
     # replace boolean arrays with sequences of integers.
     newElem = []
     IndexErrorMsg=\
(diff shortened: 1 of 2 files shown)


jswhit:
It's not clear to me what your changes actually do - it seems like they shouldn't change anything.  Can you explain what exactly is being done differently?

fredrik-1:
The first for loop just count the ellipsis and give the correct error message if two or more ellipses are used. It is needed because 
`i += 1+nDims-len(elem)`
might give the wrong index if there are more than one ellipsis. This might result in an "integer index exceeds dimension size" error instead of an "At most one ellipsis allowed in a slicing expression" error.

The behavior of the second for loop is identical but it avoid the variable i from the enumerate that is not used anyway and I think that it looks better with the update at the end. 

I get the wrong error message in master in the case when I test for example 
`var['var1'][...,range(5,8),...]=np.ones((2,3,5,20))` instead of 
`var['var1'][:, range(5,8),...]=np.ones((2,3,5,20))`
and the dimensions are (2, 15, 5, 20).
range(5, 8) are compared to the dimension of size 5 instead of the one of size 15 that was intended (but an index with two ellipsis is of course ambiguous)

jswhit:
OK, thanks for the explanation.  Could you add a test in tst_utils.py?

fredrik-1:
I can try to do it.

fredrik-1:
I wrote some general tests for ellipsis and special tests for the error message issue explained above. 

I also added a couple of `put=True` that I think where missing. 

I also realized that the datashape sequence is only used in utils.py for unlimited dimensions and not for any matching dimension tests.

Issue #720, opened 2017-10-04
Add "calendar" overload to MFTime?


bekozi:
When using `MFTime` on time bounds variables, it is often the case that the variable does not have a `calendar` attribute which triggers: https://github.com/Unidata/netcdf4-python/blob/16b376340769a05628022a8318cc8963f85a3ae1/netCDF4/_netCDF4.pyx#L6309
Is there a reason the interface for `MFTime` could not accommodate a calendar overload similar to units?

jswhit:
Seems like a default value `calendar='standard'` should work.  

Pull request is welcome.

bekozi:
Thanks @jswhit. And, yes, I'll do a PR.

bekozi:
Added in PR https://github.com/Unidata/netcdf4-python/pull/721.

PR #721, opened 2017-10-07
Add optional calendar overload to MFTime

bekozi:
I decided to go with `calendar=None` as the default to handle the time bounds case with no surprises. If standard was the default, a bounds variable associated with a time centroid having a non-standard calendar would produce unexpected behavior (if the user went with the default). Let me know if you think standard should be used instead!

Addresses #720.


jswhit:
Thanks - that looks good.  Could you add a changelog entry?

Also, feel free to ignore that CLA nag message.

bekozi:
Oops! I'll add one.

jswhit:
Thanks!   Merging now...

bekozi:
Great! Thank you.

PR #722, opened 2017-10-09
maint: updated link for pypi pages

zerothi:
The subversion repository link was still present after uploading to PyPi.
Updated to the github page.


jswhit:
Thanks!

PR #724, opened 2017-10-11
unnecessary if-else removal

milos-korenciak:
I suggest this simple optimization of the code to something straight forward.


jswhit:
Nice!  merging now, thanks ...

PR #727, opened 2017-10-17
pass errno and optional args to exceptions

jhamman:
- [x] Closes https://github.com/pydata/xarray/pull/1622 
- [x] Tests added / passed
- [x] Passes ``git diff upstream/master | flake8 --diff``

We're finding that the current "FileNotFoundError" handling in netCDF4 is not sufficiently informative (see above xarray issue). 

```
In [1]: from netCDF4 import Dataset

In [2]: Dataset('not a file')
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
<ipython-input-2-b0f2bfa73c00> in <module>()
----> 1 Dataset('not a file')

netCDF4/_netCDF4.pyx in netCDF4._netCDF4.Dataset.__init__()

netCDF4/_netCDF4.pyx in netCDF4._netCDF4._ensure_nc_success()

OSError: No such file or directory
```

Where we'd ideally raise a more informative error 
```Python-traceback
FileNotFoundError: No such file or directory: "not a file"
```

This PR does two things:

1) it passes the netCDF error code to each Error Class. This is particularly important for IOErrors but also adds some additional information to the other error classes.
2) allows a series of optional positional arguments (e.g. filename) that the error class can use.

cc @shoyer

### DIFF
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -1628,10 +1628,14 @@ cdef _get_vars(group):
         free(varids) # free pointer holding variable ids.
     return variables
 
-cdef _ensure_nc_success(ierr, err_cls=RuntimeError):
+cdef _ensure_nc_success(ierr, err_cls=RuntimeError, filename=None):
     # print netcdf error message, raise error.
     if ierr != NC_NOERR:
-        raise err_cls((<char *>nc_strerror(ierr)).decode('ascii'))
+        err_str = (<char *>nc_strerror(ierr)).decode('ascii')
+        if isinstance(err_cls, EnvironmentError):
+            raise err_cls(ierr, err_str, filename)
+        else:
+            raise err_cls(err_str)
 
 # these are class attributes that
 # only exist at the python level (not in the netCDF file).
@@ -2008,7 +2012,7 @@ references to the parent Dataset or Group.
         else:
             raise ValueError("mode must be 'w', 'r', 'a' or 'r+', got '%s'" % mode)
 
-        _ensure_nc_success(ierr, IOError)
+        _ensure_nc_success(ierr, err_cls=IOError, filename=path)
 
         # data model and file format attributes
         self.data_model = _get_format(grpid)
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -1632,7 +1632,7 @@ cdef _ensure_nc_success(ierr, err_cls=RuntimeError, filename=None):
     # print netcdf error message, raise error.
     if ierr != NC_NOERR:
         err_str = (<char *>nc_strerror(ierr)).decode('ascii')
-        if isinstance(err_cls, EnvironmentError):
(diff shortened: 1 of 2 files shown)


netCDF4/_netCDF4.pyx
```diff
@@ -1628,10 +1628,10 @@ cdef _get_vars(group):
         free(varids) # free pointer holding variable ids.
     return variables
 
-cdef _ensure_nc_success(ierr, err_cls=RuntimeError):
+cdef _ensure_nc_success(ierr, *args, err_cls=RuntimeError):
     # print netcdf error message, raise error.
     if ierr != NC_NOERR:
-        raise err_cls((<char *>nc_strerror(ierr)).decode('ascii'))
```
shoyer: DIffernet error classes use different constructors — it’s not always valid to pass an error number as the first argument. I think this function needs a path specifically for `EnvironmentError`.

jhamman:
@shoyer / @jswhit - do we want some additional test coverage on this? Something like:

```Python
    def test_no_such_file_raises(self):
        fname = 'not_a_nc_file.nc'
        try:
            netCDF4.Dataset(fname, 'r')
        except IOError as e:
            assert e.errno == errno.ENOENT
            assert e.filename == fname
            assert 'No such' in e.strerror
```

shoyer:
Yes, that would probably be nice to have.

jswhit:
travis tests are failing

jhamman:
@jswhit - give me a few minutes. I'm investigating. All the tests are passing locally so I need to sort out what's happening on Travis.

@shoyer - Here's something odd that I've never run into. My previous use of `isinstance` was causing some odd problems. 

```Python
In [1]: isinstance(IOError, EnvironmentError)
Out[1]: False

In [2]: isinstance(OSError, EnvironmentError)
Out[2]: False

In [3]: IOError is EnvironmentError
Out[3]: True

In [4]: OSError is EnvironmentError
Out[4]: True
```

shoyer:
@jhamman You need `issubclass` here, not `isinstance`.

jhamman:
@jswhit - all tests are passing.

jhamman:
Another thought I'll document here. It may be a good idea at some point to generalize this error function a bit more:

```Python
import errno

class NetcdfError(RuntimeError):
    pass

cdef _ensure_nc_success(ierr, filename=None):
    # print netcdf error message, raise error.
    if ierr != NC_NOER0:
        err_str = (<char *>nc_strerror(ierr)).decode('ascii')
        if ierr in errno.errorcode:
            raise EnvironmentError(ierr, err_str, filename)
        else:
            raise NetcdfError(err_str)   # perhaps with some more information...
```

If NetCDF-C returns an `errno` that exists in `errno.h`, it may always make sense to raise an environment error and let Python determine the best exception to present. For error codes not in `errno.h` (likely sourced from NetCDF-C, you could raise a more descriptive `NetcdfError`.

jswhit:
Merging now...

PR #732, opened 2017-10-23
allow _FillValue to be set using fill_value createVariable kwarg for vlen string vars

jswhit:


test/tst_masked5.py
```diff
@@ -41,13 +44,20 @@ def test_scaled(self):
...
+        v2 = f.variables["v2"]
         self.assertTrue(isinstance(v[:], ma.core.MaskedArray))
         assert_array_equal(v[:], self.v_ma)
         assert_array_equal(v[2],self.v[2]) # issue #624.
         v.set_auto_mask(False)
         self.assertTrue(isinstance(v[:], np.ndarray))
         assert_array_equal(v[:], self.v)
 
+        # issue 730
+        assert (v2[0]=='first')
```
shoyer: This should probably be `u'first'` for it to pass on both python 2/3. The rule of thumb is to always use bytes/unicode literals (`b'first'` and `u'first'`) unless you want native strings.

jswhit:
Looks like the travis tests are failing because they are using an older version of the netcdf-c library.  I'll disable this test if the library version is < 4.4

Issue #713, opened 2017-09-21
CDF5 support in netCDF will be optional with next release.


WardF:
[See this netcdf-c issue for more information](https://github.com/Unidata/netcdf-c/issues/484).

Starting with the upcoming 4.5.0-rc3 and moving forward, CDF5 support will be a configure time option.  There are currently a couple of issues with it that are blocking the 4.5.0-rc3 release, and adding a way to disable support has eased the logistics of getting the release out while addressing the issue.

`nc-config` has a new option, `nc-config --has-cdf5` which will return whether or not libnetcdf supports CDF5.   Let me know if there's anything I can do to help make this easy for the python library to adapt to.

jswhit:
Right now the python install script checks to see if `netcdf.h` has `#define NC_FORMAT_64BIT_DATA` in it - if so, CDF5 support is enabled.   Will this approach still work in 4.5.0?

WardF:
It will not.  We also distribute `netcdf_meta.h`, I can add something in there along the lines of `#define HAS_CDF5`, let me take a look at it.

jswhit:
Just tried running with `v4.5.0-release-branch` using `--disable-cdf5` - it compiles fine, but as expected `tst_cdf5.py` fails.  I don't think this is too big a deal though.

WardF:
Sounds good; it's an expected failure of course, and if it isn't a problem for the netcdf4-python package, it isn't a problem for us.  I wanted to say something ahead of time, however. I'm starting to prep the rc3 release today.

jswhit:
If you can add HAS_CDF5 in netcdf_meta.h that would be helpful - that way I can disable the test if it is set.

WardF:
Done; currently in the `v4.5.0-release-branch` but will be migrated out to master shortly.

ArchangeGabriel:
@jswhit Can you disable the test if NetCDF is compiled without CDF5? This is the only test failure we see in our builds here, and we would like to avoid having to add a special case handling for this in our build process. ;)

jswhit:
It's done it github master

ArchangeGabriel:
How is `__has_cdf5_format__` determined? Because 1.3.1 still fails here:
```
======================================================================
ERROR: runTest (tst_cdf5.test_cdf5)
testing NETCDF3_64BIT_DATA format (CDF-5)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #736, opened 2017-11-01
fixed bug in detection of CDF5 library support (issue #713)

jswhit:


ArchangeGabriel:
Why are there two variables: `has_cdf5` and `has_cdf5_format`? Seems strange to me.

ArchangeGabriel:
Let’s be more specific: with the current version of the code, it still fails. But if I change `has_cdf5` to `has_cdf5_format` then it works.

setup.py
```diff
@@ -81,7 +81,7 @@ def check_api(inc_dirs):
             has_cdf5 = False
```
ArchangeGabriel: This line looks problematic to me. If I build against 4.4.1.1 that has CDF5, the file exists, but not the #define. So it would be erroneously false.

ArchangeGabriel:
I think the logic is a bit flawed. AFAIK, they are four cases to be considered:
1. NetCDF < 4.4, then no CDF5. In this case we won’t find `#define NC_FORMAT_64BIT_DATA`, and `netcdf_meta.h` (if it even exists) won’t have a `#define NC_HAS_CDF5` line, so we’re good.
2. NetCDF = 4.4.x. Then CDF5 is mandatory, an we find it. However, we already have a `netcdf_meta.h` file, but it does not contain a `#define NC_HAS_CDF5` line. Which means the current code turns it off (well, not exactly because it turns `had_cdf5` off, so nothing happens).
3. NetCDF = 4.5.x without CDF5. The code works except that we don’t disable the test because of `has_cdf5` vs `has_cdf5_format`.
4. NetCDF = 4.5.x with CDF5. The code works.

So up to https://github.com/Unidata/netcdf4-python/blob/ab1128ed3762fb6ef03fd95ab661dc96b2a1dead/setup.py#L77 the code is fine.
Then I would remove [this line](https://github.com/Unidata/netcdf4-python/blob/ab1128ed3762fb6ef03fd95ab661dc96b2a1dead/setup.py#L81) because I don’t see the point of it (either you have the line and everything is said, or you don’t and having this file means nothing) and change `has_cdf5` to `has_cdf5_format` in https://github.com/Unidata/netcdf4-python/blob/ab1128ed3762fb6ef03fd95ab661dc96b2a1dead/setup.py#L84.

jswhit:
OK, should be fixed now.

setup.py
```diff
@@ -484,6 +483,11 @@ def _populate_hdf5_info(dirstosearch, inc_dirs, libs, lib_dirs):
     # this determines whether renameGroup and filepath methods will work.
     has_rename_grp, has_nc_inq_path, has_nc_inq_format_extended, \
         has_cdf5_format, has_nc_open_mem, has_nc_par = check_api(inc_dirs)
+    # for netcdf 4.4.x CDF5 format is always enabled.
+    if netcdf_lib_version > "4.4" and netcdf_lib_version < "4.5":
+        has_cdf_format = True
```
ArchangeGabriel: Typo, should be `has_cdf5_format`.

ArchangeGabriel:
OK, it works for me.

PR #739, opened 2017-11-07
remove version restrictions on libnetcdf and hdf5 in conda recipe

jswhit:


conda.recipe/meta.yaml
```diff
@@ -19,16 +19,16 @@ requirements:
     - python
     - setuptools
     - cython
-    - numpy x.x
```
ocefpaf: This is a special syntax that will test against the numpy version specified in https://github.com/Unidata/netcdf4-python/blob/master/appveyor.yml#L15

Because AppVeyor is quite slow we were testing only one, but someone could create a test matrix from it.

jswhit: OK, added that back in.

conda.recipe/meta.yaml
```diff
@@ -19,16 +19,16 @@ requirements:
...
-    - hdf5 1.8.17|1.8.17.*
-    - libnetcdf 4.4.*
+    - hdf5 
+    - libnetcdf 
   run:
     - python
     - setuptools
-    - numpy x.x
-    - hdf5 1.8.17|1.8.17.*
-    - libnetcdf 4.4.*
```
ocefpaf: I would still recommend to keep the pinning b/c testing against latest can lead to breakages that are not associated with netCDF4-python code. The only update I would do is `hdf5 1.8.18|1.8.18.*`

Note that conda-forge does have `libnetcdf 4.5.0` and `hdf 1.10` but we are not using them to build extensions yet.

jswhit: but I would expect netcdf4-python to work with the latest releases of hdf5 and netcdf-c, and if it doesn't I would like to know about it.

ocefpaf: It should work. We (conda-forge) are not using it b/c we cannot user `hdf5 1.10` with `libnetcdf 4.4.1.1` and `libnetcdf 4.5.0` does not build on vc 9 (Python 2.7).

TL;DR we should wither patch libnetcdf 4.5.0 to build with vc 9 or drop Python 2 in order to upgrade here.

PR #747, opened 2017-12-13
more pythonic code

milos-korenciak:
More pythonic equivalent without unnecessary local variable; this uses generator (surely c-optimized for-loop) and sum (c optimized?)

### DIFF
--- a/netCDF4/utils.py
+++ b/netCDF4/utils.py
@@ -191,7 +191,10 @@ def _StartCountStride(elem, shape, dimensions=None, grp=None, datashape=None,\
         elem = [elem]
     
     # ensure there is at most 1 ellipse
-    if elem.count(Ellipsis) > 1:
+    #  we cannot use elem.count(Ellipsis), as with fancy indexing would occur
+    #  np.array() == Ellipsis which gives ValueError: The truth value of an 
+    #  array with more than one element is ambiguous. Use a.any() or a.all()
+    if sum(1 for e in elem if e is Ellipsis) > 1:
         raise IndexError("At most one ellipsis allowed in a slicing expression")
     
     # replace boolean arrays with sequences of integers.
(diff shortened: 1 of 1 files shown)


milos-korenciak:
Fixing code to be more pythonic.
Maybe it can be divided into more lines, but it seems to me still perfectly readable. (Maybe I am wrong.)

netCDF4/utils.py
```diff
@@ -190,12 +190,10 @@ def _StartCountStride(elem, shape, dimensions=None, grp=None, datashape=None,\
...
         elem = [elem]
     
-    hasEllipsis=False
-    for e in elem:
-        if type(e)==type(Ellipsis):
-            if hasEllipsis:
-                raise IndexError("At most one ellipsis allowed in a slicing expression")
-            hasEllipsis=True
+    # ensure there is at most 1 ellipse
+    if sum((1 for e in elem if type(e) == type(Ellipsis))) > 1:
```
shoyer: You can omit the extra `()` inside sum.

Also, there is only one `Ellipsis` object (it's a singleton), so `e is Ellipsis` would suffice.

jswhit:
Maybe it's just me, but I find the original code more readable (but definitely more verbose).

willirath:
> Maybe it's just me, but I find the original code more readable (but definitely more verbose).

I agree.  What about:
```python
if elem.count(Ellipsis) > 1:
    raise IndexError("At most one ellipsis allowed in a slicing expression")
```

netCDF4/utils.py
dopplershift: You also don't need the extra inner set of parenthesis.

milos-korenciak:
I definitely love willirath solution :+1:

milos-korenciak:
So we canot use elem.count as it uses == for comparison. I.e. when used fancy indexing (= numpy array used as indexes) there occurs internal comparison "numpy.array == Ellipsis", which gives:
"ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()"
So, back to shoyer's solution.

jswhit:
looks good, merging now - thanks!

PR #756, opened 2018-01-31
remove netcdftime (use separate project)

jswhit:


jhamman:
Thanks @jswhit - I suppose this is going to be blocked until we get a pypi / conda-forge build ready for [netcdftime](https://github.com/Unidata/netcdftime).

PR #789, opened 2018-03-24
nc3tonc4 handles _FillValue incorrectly

kushim6371:
nc3tonc4 handles _FillValue incorrectly when unpacking shorts.

This update uses two variables, fillvalue3 and fillvalue4, for the packed shorts in nc3file and the unpacked floats in nc4file, respectively.

Since maskandscale is set to on by default, the manual "mask and scale" is unnecessary and removed in this update.


jswhit:
Thanks!

PR #793, opened 2018-04-17
netCDF4/utils.py bug fixed to call _getgrp function

hfattahi:
A bug fixed to call _getgrp function inside netCDF4/utils.py
Before this fix ncinfo does not provide group information when -g option is requested.


jswhit:
good catch, thanks!

Issue #809, opened 2018-06-22
Use of masked arrays comes with huge performance drop


fmaussion:
The change included in https://github.com/Unidata/netcdf4-python/pull/787 made our test suite run in two hours instead of a couple of minutes.

I am not well placed to judge whether this was a good change or not, but I spent several hours in finding out what was going on. <s>It would be great if these changes were documented somewhere, i.e. in a "what's new section" or similar.</s> EDIT: just discovered the ``Changelog``. Still, a link to it in the html documentation would be of great help.

Right now this makes v1.4 unusable for us: is there a way to ``set_auto_mask`` to False globally?

fmaussion:
An example with indexing:

```python
import numpy as np
import time

a = np.linspace(0, 100, 2000)
ma = np.ma.array(a)

ids = np.arange(1000, 1500)

%timeit a[ids]
631 ns ± 13.3 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

%timeit ma[ids]
14.3 µs ± 188 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

davidhassell:
I, too, have been anecdotally concerned about this aspect of performance, but haven't got round to properly testing it in my code; so thanks for raising this.

An option to produce the pre-1.4 behaviour- i.e. return non-masked numpy arrays when there are no elements of fillValue, and masked numpy arrays when there are - would be great for me.

Something like `set_dynamic_mask`, perhaps? This would default to `False`, meaning the default 1.4 behaviour, or could be set to `True` for the pre-1.4 behaviour.

Thanks, David

davidhassell:
... if this is of interest, I'd be happy to create pull request with some new code ....

fmaussion:
I can't imagine that the majority of users will be happy with the current 1.4 behavior. Masked arrays are a not well understood part of numpy, beginners have troubles with them, and they come with a huge performance loss (factor 20 in the indexing example above). What's going to happen is that most users will set the masking to false per default, or wrap their own ``Dataset`` with this as default (this what we will probably do instead of enforcing netCDF v1.3).

I wonder if the path isn't to go through a painful (but rewarding on long term) deprecation of the pre 1.4 behavior: set ``auto_mask`` to ``False`` per default...

jswhit:
You can turn auto-masking off globally using the set_auto_mask Dataset method.

I think there are only two choices that make sense - always return masked arrays or always return numpy arrays.

[...]

Issue #808, opened 2018-06-21
Empty mask being returned as full mask array 


davidhassell:
Hello,

Since `v1.4`, the code at https://github.com/Unidata/netcdf4-python/blob/master/netCDF4/_netCDF4.pyx#L4293-L4298 suggests that when there are no masked elements, the mask of the variable's numpy array should be of type `numpy.bool_`, via the setting of `mask=False`. 

This is not what I'm seeing, though:

```
import netCDF4, numpy

print netCDF4.__version__, numpy.__version__

n = netCDF4.Dataset('test1.nc', 'w')

x = n.createDimension('x', 4)
y = n.createDimension('y', 3)

var = n.createVariable('var', 'f8', ('y', 'x'))
var.units = "metre"
var[...] = numpy.arange(12)

n.close()

n2 = netCDF4.Dataset('test1.nc', 'r')
print repr(n2.variables['var'][...])
```
Outputs:
```
1.4.0 1.13.3
masked_array(data =
 [[0.0 1.0 2.0 3.0]
 [4.0 5.0 6.0 7.0]
 [8.0 9.0 10.0 11.0]],
             mask =
 [[False False False False]
 [False False False False]
 [False False False False]],
       fill_value = 9.96920996839e+36)
```
i.e. a full mask of type `numpy.ndarray` full of Falses.

Compared with:
```
In [2]: numpy.ma.arange(12).reshape(3, 4)
Out[2]: 
masked_array(data =
 [[ 0  1  2  3]
 [ 4  5  6  7]
 [ 8  9 10 11]],
             mask =
 False,
       fill_value = 999999)
```
Am I expecting correctly? and if so, am I doing something wrong, or is the code not behaving?

Many thanks for you help,

David

jswhit:
I would have expected the array to have a boolean False mask, not a ndarray mask full of Falses.

However, it looks like if I leave out ``mask=False``, the mask will be type `np.bool_`.  Is there a particular reason to prefer this behavior?  If so, I'll go ahead and make a pull request.

davidhassell:
The reason is to save on memory - a float32 array with a full mask of `False`s is 25% larger than it needs to be. Also performance - on my laptop it is 40x faster to test for any masked points if the mask is `numpy.bool_`:

davidhassell:
```
In [20]: z = numpy.ones((1000000,))

In [21]: a = numpy.ma.array(z, mask=False)

In [22]: b = numpy.ma.array(z)

[...]

PR #816, opened 2018-07-16
new methods to optionally re-enable old numpy array type behaviour (issue #809)

davidhassell:


jswhit:
Looks good, thanks @davidhassell

akrherz:
Just a comment from the peanut gallery here and I know API design is hard, but the name `set_auto_array_type` does not seem intuitive to me.  This setting is controlling whether variables should always be numpy masked arrays or not.  Currently `set_auto_mask(self, True_or_False)` exists, it seems like a companion to this would be `set_always_mask(self, True_or_False)`  ?

Just my unqualified two cents here.

davidhassell:
@jswhit, are you happy to rename the methods, as per @akrherz's suggestion?

jswhit:
Not quite clear on what `set_always_mask` would do.  Would it be `True` but default, and to enable the old behavior it would be set to `False`?

jswhit:
as per the discussion in issue #809, please go ahead and change
 
```python
            # issue #785: always return masked array, if no values masked
            # set mask=False.
            data = ma.masked_array(data,mask=False,fill_value=fill_value)
```

to 

```python
            # issue #785: always return masked array by default, if no values masked
            data = ma.masked_array(data,fill_value=fill_value)
```

davidhassell:
Yes - The logic needs reversing when we rename to `set_auto_array_type` to `set_always_mask`. The default would be True, signifying that masked arrays are returned in every instance. I'll add this.

jswhit:
Looks like tst_multifile.py is failing somewhere in cftime.  Restoring the 'mask=False' fixes it.

davidhassell:
I suggest, then, that we restore `mask=False` here and revisit it in #808 (i.e. see if we can fix the downstream issues in cftime as pull request on that ticket (which I'm happy to look at, if you like)). I'll put `mask=False` back in this ticket's PR. Thanks

PR #818, opened 2018-07-20
create masked array with empty mask, instead of mask full of Falses

jswhit:
(issue #808).  Also bump version number to 1.4.1


jswhit:
parts of tst_multifile.py and tst_multifile2.py have been disabled for cftime < 1.0.1 (the next release which will include a fix that will allow those tests to pass)

Issue #824, opened 2018-08-08
How to obtain paths for dimensions linked to variables


PaulHuwe:
I have an nc4 file with groups (and subgroups) containing variables and dimensions at various levels within the group structure. When I try to obtain the dimensions that a specific variable maps, netcdf4-python only provides the pathless dimension name (and size in a tuple). Is there a way to obtain the dimension name with path, or a link to the dimension object (which I could then use the .group() method on)? 

An example:
netcdf file {
dimensions:
        airs_idx = 143 ;
...
group: AIRS {
  dimensions:
        Cloud = 2 ;
...
  group: RetStd{
    variables:
        float CldFrcStd(airs_idx, Cloud) ;
...
    }
  }
}

When I perform the following in python, I get:
>>> f['/AIRS/RetStd/CldFrcStd'].dimensions
(u'airs_idx', u'Cloud')

When I would like to know that the dimensions are:
(u'/airs_idx', u'/AIRS/Cloud')

Thanks

jswhit:
From https://www.unidata.ucar.edu/software/netcdf/docs/netcdf_data_set_components.html

"Dimensions are scoped such that they can be seen in all descendant groups. That is, dimensions can be shared between variables in different groups, if they are defined in a parent group."

Given this, I'm not sure it makes sense to assign a path to a dimension.  Internally, `netCDF4.utils._find_dim` is used to find a dimension instance the the Group and name, by searching the current group and then traversing it's parents.  I suppose we could expose this function

jswhit:
BTW, here is the code for _find_dim, which actually is accessible, just not documented.

```python
def _find_dim(grp, dimname):
    # find Dimension instance given group and name.
    # look in current group, and parents.
    group = grp
    dim = None
    while 1:
        try:
            dim = group.dimensions[dimname]
            break
        except:
            try:
                group = group.parent
            except:
                raise ValueError("cannot find dimension %s in this group or parent groups" % dimname)
    return dim
```

jswhit:
Perhaps what is needed is a variable method to retrieve the actual dimension objects, not just their names?

PaulHuwe:
I think such a method would both be useful and consistent with the rest of the package.

jswhit:
Pull request #825  adds a `get_dims` Variable method that returns a tuple of dimension objects associated with the Variable.

jswhit:
pull request #825 merged, closing now

PR #825, opened 2018-08-11
add get_dims Variable method (issue #824)

jswhit:


netCDF4/_netCDF4.pyx
```diff
@@ -5124,6 +5124,19 @@ open for parallel access.
...
+**`get_dims(self)`**
+
+return a tuple of `netCDF4.Dimension` instances associated with this 
+`netCDF4.Variable.
+        """
+        dims = ()
+        for dimname in self.dimensions:
+            dim = _find_dim(self._grp, dimname)
+            dims += (dim,)
+        return dims
```
jhamman: I think this could be much more concisely written as

```Python
return (_find_dim(self._grp, dim) for dim in self.dimensions)
```

jswhit: good idea, thanks @jhamman.  Needs to be wrapped with tuple() though.

PR #831, opened 2018-08-17
Update appveyor

ocefpaf:
@jswhit we are still using conda to get the dependencies b/c it is convenient, but I'm not using `conda-build` for building and running the tests in a similar fashion we did for `cftime`. The main drive for this is to reduce the boilerplate, avoid confusion (people may thing the conda package is produce from the test recipe), and to avoid problems when conda-build returns a green status when the recipe build actually failed.


jswhit:
This simplifies things a lot @ocefpaf, thanks.

PR #832, opened 2018-08-17
enable travis test for 3.7, drop 3.5

jswhit:


jswhit:
no python 3.7 in travis yet

https://github.com/travis-ci/travis-ci/issues/9815

jswhit:
specifying dist: xenial allows 3.7 to work.  However, the MPI test won't work with xenial, so I'm still using trusty for that one.

PR #864, opened 2018-12-20
fix so test passes with cftime master

jswhit:
Also fixes failing tst_diskless.py with netcdf-c 4.6.2 (requires additional NC_PERSIST flag to nc_open/nc_create)

### DIFF
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -1239,7 +1239,7 @@ NC_DISKLESS = 0x0008
 if __netcdf4libversion__[0:5] >= "4.6.2":
     NC_PERSIST = 0x4000
 else:  # prior to 4.6.2 this flag doesn't work, so make the same as NC_DISKLESS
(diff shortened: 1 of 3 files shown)


netCDF4/_netCDF4.pyx
```diff
@@ -1235,6 +1235,11 @@ is_native_big = numpy.dtype('>f4').byteorder == '='
 # hard code these here, instead of importing from netcdf.h
 # so it will compile with versions <= 4.2.
 NC_DISKLESS = 0x0008
+# introduced in 4.6.2
+if __netcdf4libversion__[0:5] >= "4.6.2":
+    NC_PERSIST = 0x4000
+else:  # prior to 4.6.2 this flag doesn't work, so make the same as NC_DISKLESS
+    NC_PERSIST = 0x0008
```
dopplershift: So why not replace hard-coded `0x0008` with `NC_DISKLESS`.

jswhit: Good point - done

Issue #865, opened 2019-01-23
Accessing the memory of a Dataset created as "diskless"


nmassey001:
When creating a Dataset with "diskless=True":

`ncd = netCDF4.Dataset(filepath, mode='w', diskless=True)`

It doesn't seem possible to access the buffer for the Dataset.

I've looked in the source code and the call to the underlying C library is:
` ...
                elif diskless:
                    if persist:
                        ierr = nc_create(path, NC_WRITE | NC_CLOBBER |
                                NC_DISKLESS | NC_PERSIST, &grpid)
                    else:
                        ierr = nc_create(path, NC_CLOBBER | NC_DISKLESS , &grpid)
...
`

The Python library doesn't seem to record a reference to the buffer created by nc_create.
Would it be possible to be able to expose this buffer?

jswhit:
As is stands now you can create a netcdf file from an existing python memory buffer, but you can't create a reference to the memory of an existing in-memory Dataset.  The former is done by using the INMEMORY capability of the C lib, the latter using the DISKLESS capability, which are separate (https://www.unidata.ucar.edu/software/netcdf/docs_rc/md__Users_wfisher_Desktop_gitprojects_netcdf-c_docs_inmemory.html).  However, from reading that page again  it does look like it may be possible to re-access the buffer of a INMEMORY file after data has been written to it using the `nc_close_memio` function.

jswhit:
Just curious, what's your use-case?

jswhit:
@DennisHeimbigner - is it possible to access the memory buffer of a file created with the NC_DISKLESS flag?  Or, is it preferable to create an in-memory file using `nc_create_mem` and then access the memory buffer from `nc_close_memio`?

DennisHeimbigner:
A case I had not considered. Looking at the code, it appears that you must
use nc_create_mem in order to extract the memory content.

jswhit:
Right now, `netCDF4.Dataset` creates a (read-only) in-memory file if` mode='r'` and a memory buffer is passed in via the `memory` kwarg.  This could be extended to `mode='w'`, and a memory buffer could be returned when the `close` method is invoked.  In this case, I guess the `memory` kwarg would just be used to indicate an in-memory file is desired (if it is not equal to `None`).  Would this satisfy your use case?

DennisHeimbigner:
You probably need either another close operation or a keyword for existing close
to indicate that you want the in-memory data.

jswhit:
[...]

PR #866, opened 2019-01-26
access bytes written to an in-memory Dataset (issue #865)

jswhit:
if `mode='w'` and `memory` kwarg is set, an in-memory `Dataset`  will be created and the bytes written will be returned by `Dataset.close()`

### DIFF
--- a/include/membuf.pyx
+++ b/include/membuf.pyx
@@ -33,10 +33,12 @@ cdef class MemBuf:
     cdef const void *get_mem(self):
         return self.p
 
-    def __getbuffer__(self, Py_buffer *view, int flags):
-        PyBuffer_FillInfo(view, self, <void *>self.p, self.l, 1, flags)
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
+        cdef int ret,readonly
+        readonly=1
+        ret=PyBuffer_FillInfo(buffer, self, <void *>self.p, self.l, readonly, flags)
 
-    def __releasebuffer__(self, Py_buffer *view):
+    def __releasebuffer__(self, Py_buffer *buffer):
         #PyBuffer_Release(view) 
         pass
 
--- a/include/membuf.pyx
+++ b/include/membuf.pyx
@@ -7,60 +7,48 @@ from cpython.bytes cimport PyBytes_FromStringAndSize
 from libc.stdlib cimport free
 from libc.stdint cimport uintptr_t
 
-ctypedef void dealloc_callback(const void *p, size_t l, void *arg)
+ctypedef void dealloc_callback(const void *memory, size_t size, void *arg)
 
-cdef void free_buf(const void *p, size_t l, void *arg):
-    free(<void *>p)
+cdef void free_buf(const void *memory, size_t size, void *arg):
+    free(<void *>memory)
 
 # this is the function used to create a memory view from
 # a raw pointer.
 # Only this function is intended to be used from external
 # cython code.
-cdef memview_fromptr(void *p, size_t l):
-    assert p!=NULL, "invalid NULL buffer pointer"
-    return memoryview( MemBuf_init(p, l, &free_buf, NULL) )
+cdef memview_fromptr(void *memory, size_t size):
+    # memory is malloced void pointer, size is number of bytes allocated
+    if memory==NULL:
+        raise MemoryError('no memory allocated to pointer')
+    return memoryview( MemBuf_init(memory, size, &free_buf, NULL) )
 
 cdef class _MemBuf:
-    cdef const void *p
-    cdef size_t l
-    cdef dealloc_callback *dealloc_cb_p
+    cdef const void *memory
+    cdef size_t size
+    cdef dealloc_callback *dealloc_cb
     cdef void *dealloc_cb_arg
-
-    def __len__(self):
-        return self.l
-
-    def __repr__(self):
-        return "_MemBuf(%#x)" % (<uintptr_t> self.p)
-
-    cdef const void *get_mem(self):
-        return self.p
 
     def __getbuffer__(self, Py_buffer *buf, int flags):
         cdef int ret,readonly
         readonly=1
-        ret=PyBuffer_FillInfo(buf, self, <void *>self.p, self.l, readonly, flags)
+        ret=PyBuffer_FillInfo(buf, self, <void *>self.memory, self.size, readonly, flags)
 
     def __releasebuffer__(self, Py_buffer *buf):
         # why doesn't this do anything??
         pass
 
     def __dealloc__(self):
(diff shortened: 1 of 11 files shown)


jswhit:
Docs still need to be updated.

netCDF4/_netCDF4.pyx
```diff
@@ -2273,14 +2289,36 @@ version 4.1.2 or higher of the netcdf C lib, and rebuild netcdf4-python."""
...
+
+        self._isopen = 0 # indicates file already closed, checked by __dealloc__
+
+        # Only release buffer if close succeeded
+        # per impl of PyBuffer_Release: https://github.com/python/cpython/blob/master/Objects/abstract.c#L667
+        # view.obj is checked, ref on obj is decremented and obj will be null'd out
+        PyBuffer_Release(&self._buffer)
+
+        # return bytes representing in-memory dataset
+        return PyBytes_FromStringAndSize(<char *>memio.memory, memio.size)
```
shoyer: Two downsides of this approach:
1. You have to close a dataset to copy it's memory contents -- but you might like to copy the data without closing the dataset just yet.
2. This will always make a copy of the Dataset's memory into a Python `bytes` object. There's no way to read data without copying all of it.

Instead, what about exposing the contents of `self._buffer` via Python's Buffer interface? My inclination would be to write a `netCDF4.Dataset.buffer` property that returns an object satisfying CPython's buffer protocol. I think this would be [pretty easy in Cython](https://cython.readthedocs.io/en/latest/src/userguide/buffer.html) and would no-copy access in Python, e.g.,by creating a `BytesIO` object like `io.BytesIO(dataset.buffer)`.

netCDF4/_netCDF4.pyx
```diff
@@ -2273,14 +2302,39 @@ version 4.1.2 or higher of the netcdf C lib, and rebuild netcdf4-python."""
...
+
+            self._isopen = 0 # indicates file already closed, checked by __dealloc__
+
+            # Only release buffer if close succeeded
+            # per impl of PyBuffer_Release: https://github.com/python/cpython/blob/master/Objects/abstract.c#L667
+            # view.obj is checked, ref on obj is decremented and obj will be null'd out
+            PyBuffer_Release(&self._buffer)
+
+            # return bytes representing in-memory dataset
+            return PyBytes_FromStringAndSize(<char *>memio.memory, memio.size)
```
shoyer: I think you may have a memory leak here (unless Cython manages this automatically somehow):
"WARNING, the returned memory is owned by the caller and so the caller is responsible for calling free() on that returned memory."
https://www.unidata.ucar.edu/software/netcdf/docs/md__Users_wfisher_Desktop_v4_86_82_netcdf-c_docs_inmemory.html

jswhit: you're right - that should be fixed now.

netCDF4/_netCDF4.pyx
jswhit: You have a point - will think about this some more.

shoyer: The other reason why I like the attribute is that you can still use contextmanger for the file, e.g.,
```python
with netCDF4.Dataset('somewhere', mode='w', diskless=True) as ds:
    ...
bytes(ds.buffer)  # could also make this a method, e.g., get_buffer() or get_file_image() (like pytables)
```

One minor correction: clean-up should happen in the `__dealloc__` or `__del__` method of the object holding the buffer, in case it has a longer lifetime than the Dataset.

It looks like this StackOverflow post has a good description of how to make these objects:
https://stackoverflow.com/a/28166272/809705

jswhit: Working on a solution that returns a memoryview object instead of a bytes object (without making a copy).  See https://stackoverflow.com/questions/28160359/how-to-wrap-a-c-pointer-and-length-in-a-new-style-buffer-object-in-cython

Don't really see the advantage of keeping the memoryview as an attribute of the Dataset object - seems counterintutive to keep using the Dataset after it's closed.

jswhit: Sorry I didn't fully read your comment above - you already referenced that stackoverflow and pointed out that keeping the buffer as an attribute lets you use the context manager.

include/membuf.pyx
```diff
@@ -0,0 +1,63 @@
...
+
+    def __releasebuffer__(self, Py_buffer *view):
+        #PyBuffer_Release(view) 
+        pass
+
+    def __dealloc__(self):
+        if self.dealloc_cb_p != NULL:
+            self.dealloc_cb_p(self.p, self.l, self.dealloc_cb_arg)
+
+    def tobytes(self):
```
shoyer: I don't know if you really need this -- you can also just write `bytes(buffer)` directly

netCDF4/_netCDF4.pyx
```diff
@@ -1060,6 +1061,62 @@ Note that there is currently no support for mapping numpy structured arrays with
...
+    >>> # used if format is NETCDF3 (ignored for NETCDF4/HDF5 files).
+    >>> nc = Dataset('inmemory.nc', mode='w',memory=1028)
+    >>> d = nc.createDimension('x',None)
+    >>> v = nc.createVariable('v',np.int32,'x')
+    >>> v[0:5] = np.arange(5)
+    >>> nc_buf = nc.close() # close returns memory buffer.
+    >>> # save nc_buf to disk, read it back in and check.
+    >>> # tobytes method of cython memory buffer converts to bytes.
+    >>> f = open('inmemory.nc', 'w')
+    >>> f.write(nc_buf.tobytes()); f.close()
```
shoyer: You can also just pass a buffer directly into the write() method.

include/membuf.pyx
jswhit: Nope - the cython MemBuf object is not a python memoryview, and bytes expects one.

netCDF4/_netCDF4.pyx
jswhit: No -that doesn't work either, probably for the same reason that bytes(buffer) doesn't.

include/membuf.pyx
jswhit: Actually, I'm wrong - bytes(buffer) does not throw an error but the result is incorrect (not the same as buffer.tobytes()).  Don't understand why.

netCDF4/_netCDF4.pyx
jswhit: actually, this does work for NETCDF3 files, but for NETCDF4 files it throws this error
    TypeError: expected a string or other character buffer object

jswhit: Dataset.close now returns a python memoryview, which can be passed directly to the write() method.  This also makes the MemBuf.tobytes() method redundant.

PR #882, opened 2019-02-24
Feature: Use variable length ncstring text attributes as default

cmarquardt:
Hi,

for some reason I need to generate netCDF files using variable length string attributes (i.e., `NCSTRING`-typed text attributes) rather than the usual `NCCHAR` character array strings. I'm aware that such string attributes can be created with `setncattr_string()`; but using different syntax for text and numerical attributes doesn't seem right to me.

On the other hand, the current implementation, if I understand correctly, is already creating `NCSTRING` attributes for unicode strings as well as for lists of strings. To me it would thus seem natural to offer an option (say at the time when a new netCDF file is created) so that all string attributes are generated as variable `NCSTRING` ones.

This pull request provides an additional optional boolean argument to `Dataset()` (named `ncstring_attrs`); if `False` (the default), data sets are created as before. If set to `True`, all text attributes use the variable length `NCSTRING` type instead.

Would that be useful?

  Christian.


jswhit:
If you want to read  the discussion that led to the current behavior, see https://github.com/Unidata/netcdf4-python/issues/529.

I'm not opposed to adding another kwarg to Dataset, although it's getting kind of cluttered already.

BTW, another hackish way to force an attribute to be NC_STRING is to append a unicode blank to it

```python
import netCDF4
nc = netCDF4.Dataset('test.nc','w',format='NETCDF4')
nc.foo = u'foo' # will be written as NC_CHAR
nc.bar = u'bar\u00a0' # append blank, will be written NC_STRING
nc.close()
```

jswhit:
Cleaned up the docstrings a little, and added a Changelog entry.

cmarquardt:
Oh thanks for cleaning up my mess!

Another option (instead of adding a kwarg to `Dataset()` would be a set of methods like `set_ncstring_attrs_on()` and `set_ncstring_attrs_off()`, or a single one like `set_ncstring_attrs(True_or_False)`. I think it was an idea also coming up in the discussion in #529 - would you prefer that? I'm happy to update the pull request...

By the way, thanks for pointing me to that discussion - I understand much better now where the complication comes from.

jswhit:
I like the idea of adding a method to keep `Dataset.__init__` less cluttered.

cmarquardt:
Would this do? If you prefer an on/off pair of methods, I'd implement that on top of the `set_ncstring_attrs()` one, making the latter an internal method.

The test only checks a use case with activating ncstring attributes immediately after creating the netCDF file. If it's called later on, only attributes created afterwards will be affected by the setting, obviously... but there is no test for that right now.

If the method is called for a subgroup, settings will only be changed for that subgroup and all its children (subgroups and variables), but not the parent group(s). This differs from `set_fill_on()`/`set_fill_off()`; personally, I wouldn't traverse the group hierarchy upwards - if a user wants to update the behaviour for the entire dataset, he/she should call the main Dataset's method IMHO.

jswhit:
That looks fine - thanks!  One small detail - why did you create a new test file (tst_atts_ncstring.py) instead of just adding a few lines to tst_atts.py?  Looks like the majority of tst_atts_ncstring.py is a copy of what's in the other file.

cmarquardt:
I think apart from the obvious laziness, I don't know of a good way to parameterize unit tests - and in the initial version of the change that would have been required because it affects all text attributes in the file. But now, with a function, enforcing NC_STRINGs can be activated at an arbitrary point in the creation of the test file...- I'll update the original test!

Thanks for pointing this out...

cmarquardt:
Hi,

test updated - and a regression fixed. tests are actually running fine in my account on travis (https://travis-ci.com/cmarquardt/netcdf4-python/builds/102403529); unfortunately I can't access the build logs on Appveyor (no "Details" link, and Travis results are not shown for some reason).

The conflict comes from the Changelog entries. I can resolve that; is an ordinary merge ok, or would you prefer a rebase?

Thanks!

jswhit:
an ordinary merge is OK, thanks.

cmarquardt:
Done... Related: Much of the discussion on the type of string attributes is contained in a few issues and merge requests - would it make sense to have a brief summary on the default behaviour in the Tutorial?

jswhit:
Looks good, merging now. Thanks @cmarquardt !  Regarding the tutorial, I don't think it's necessary - the API docs should suffice.

cmarquardt:
Fair enough..

Thanks for accepting the pull request!

PR #902, opened 2019-03-26
Feature/read shared

barronh:
Adding read shared mode.

netCDF4.Dataset currently has append shared, but no read shared. There are two circumstances that require read shared. First, read shared provides access to edits made by other programs as they become available. Second, read shared makes reading faster than read.

Shared is already available through append (as or r+s), but this leads to confusion when the user does not want append access. By enabling append, the user may accidentally make edits to a file. This creates a hesitancy to use as or r+s even when a program may be enhanced dramatically.

I tested a NetCDF3 file with:
TSTEP = 744
ROW = 299
COL = 459

With `r`, the file took 14 seconds to read a single variable (all times). With `rs`, the file took less than a second.

### DIFF
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -1998,7 +1998,7 @@ strings.
         modified. `w` means write; a new file is created, an existing file with
         the same name is deleted. `a` and `r+` mean append (in analogy with
         serial files); an existing file is opened for reading and writing.
-        Appending `s` to modes `w`, `r+` or `a` will enable unbuffered shared
+        Appending `s` to modes `r`, `w`, `r+` or `a` will enable unbuffered shared
         access to `NETCDF3_CLASSIC`, `NETCDF3_64BIT_OFFSET` or
         `NETCDF3_64BIT_DATA` formatted files.
         Unbuffered access may be useful even if you don't need shared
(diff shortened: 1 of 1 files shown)


jswhit:
DIdn't realize this would have such an impact on read speed.  Thanks for the contribution @barronh!

If you can add a Changelog entry, I'll go ahead and merge.

netCDF4/_netCDF4.pyx
```diff
@@ -2202,7 +2202,14 @@ strings.
             elif diskless:
                 ierr = nc_open(path, NC_NOWRITE | NC_DISKLESS, &grpid)
             else:
-                ierr = nc_open(path, NC_NOWRITE, &grpid)
+                if mode == 'rs':
+                    # NC_SHARE is very important for speed reading
+                    # large netcdf3 files with a record dimension.
+                    # Opening as r+s or as implies capability to
+                    # which may be inconsistent with actual access
```
jswhit: Not sure what this means - 'capability to which may be inconsistent with actual access'?
Would be useful to have a reference to this pull request in the comments.

barronh: @jswhit - I have just returned from vacation and looking at your comment. I haven't done anything with the changelog, but am happy to.

What I meant was that the append modes (a or r+) imply that the user has the capability to write. So, any properties tied to the open mode will expect write capability. If the user doesn't have permission, there is an inconsistency there.

Sorry for the delay.

barronh: @jswhit - Since you've merged it, do you still want me to fill out a log?

jswhit: No worries, I did it.

Issue #930, opened 2019-05-29
BUG: read nc file get unexpected values


Clarmy:
When I read a `.nc` file by `netCDF4` library, the result is unexpected.
The code:
```python
In [1]: import netCDF4 as nc

In [2]: pfn = './test/FY4A-_AGRI--_N_REGC_1047E_L2-_FHS-_MULT_NOM_20190523003000_20190523003416_2000M_V0001.NC'

In [3]: fileObj = nc.Dataset(pfn)

In [4]: fhsObj = fileObj.variables['FHS']

In [5]: fhs = fhsObj[:]

In [6]: fhs.max()
Out[6]: 56320

In [7]: fhs.min()
Out[7]: 2560

In [8]: set(fhs.data.flatten())
Out[8]: {2560, 10240, 12800, 15360, 25600, 38400, 39168, 46080, 56320, 65280}
```
The result shows array consist of 2560, 10240, 12800, ... ,65280.
While if I read this file by Panoply. The result is:
![image](https://user-images.githubusercontent.com/20877876/58559033-ce35b200-8254-11e9-865c-1ec87bf51a46.png)

This result is much different from `netCDF4`'s.
And I tried `xarray` to read this file:
```python
In [9]: import xarray as xr

In [10]: ncfile = xr.open_dataset(pfn)

In [11]: set(ncfile.FHS.values.flatten())
Out[11]: {10.0, 40.0, 50.0, 60.0, 100.0, 150.0, 153.0, 180.0, 220.0, 255.0}
```
This result agrees with Panoply's. And it should be right according to the file's attribute. So I guess it's a bug in `netCDF4`

version information:
`Python 3.7.1 netCDF4 1.4.2`

jswhit:
xarray uses netCDF4, so I suspect it has do with the automatic scaling of short integers into floats.  Can you attach the gzipped file to this ticket - or if it is too large post it somewhere where we can access it?

Clarmy:
Sorry, I forgot to upload example file.
[FY4A-_AGRI--_N_REGC_1047E_L2-_FHS-_MULT_NOM_20190523003000_20190523003416_2000M_V0001.zip](https://github.com/Unidata/netcdf4-python/files/3233070/FY4A-_AGRI--_N_REGC_1047E_L2-_FHS-_MULT_NOM_20190523003000_20190523003416_2000M_V0001.zip)

jswhit:
[...]

PR #931, opened 2019-05-29
fix for scaling bug when _Unsigned is set and data byteorder not native (issue #930)

jswhit:


jswhit:
needs test

jswhit:
test added in test/tst_endian.py

PR #950, opened 2019-07-08
Revise docstrings with Python 3

mwtoews:
This is primarily a refresh to the main docstrings using Python 3

- obviously `print x` becomes `print(x)`
- use "with" context managers in a few places, where it makes sense
- several errors or inconsistencies were identified and corrected

The docstrings were processed and compared using this workflow:
```bash
head -1226 netCDF4/_netCDF4.pyx > x.py
python -m doctest -v x.py | less
```
a few notes on this workflow:
- Common corrections, like turning `>>>` to `...` are corrected to avoid syntax errors
- There are many `<BLANKLINE>` after printing netCDF4 objects. Consider modifying the behaviour to not add a newline character to these objects (I'm guessing somewhere in `__repr__` methods)
- Where no output was expected, a `_ = ` was added to assign a dummy variable, which is common in Python code. However, these can be removed with the output in place.
- Group variables are sometimes underlined using escape characters, e.g. `variables(dimensions): float32 ESC[4mtempESC[0m(time,level,lat,lon)`; it would be nice to have a setting to turn these off (perhaps there already is?)

### DIFF
--- a/netCDF4/_netCDF4.pyx
+++ b/netCDF4/_netCDF4.pyx
@@ -151,7 +151,7 @@ Here's an example:
     :::python
     >>> from netCDF4 import Dataset
     >>> rootgrp = Dataset("test.nc", "w", format="NETCDF4")
-    >>> print rootgrp.data_model
+    >>> print(rootgrp.data_model)
     NETCDF4
     >>> rootgrp.close()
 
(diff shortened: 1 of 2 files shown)


jswhit:
appveyor tests are failing because they can't find packages (unrelated to the code changes in this PR). 

@ocepaf - do you have any suggestions on how to fix this?

ocefpaf:
> @ocefpaf - do you have any suggestions on how to fix this?

I'll take a look at it in a moment but this is safe to merge without it.

PR #952, opened 2019-07-09
Fix Appveyor

ocefpaf:
Hopefully just using the new strict is enough to fix this.


jswhit:
Thanks for acting on this so quickly @ocefpaf !

PR #955, opened 2019-07-10
Refactor Python 3.7+ to use dict built-in instead of OrderedDict

mwtoews:
A nice [feature of Python 3.7](https://docs.python.org/3/whatsnew/3.7.html) is that insertion order for `dict` objects is guaranteed. This means that `OrderedDict` is no longer needed to preserve this behavior.

One good benefit of using the standard `dict` is that it has a more compact string representation (e.g. `OrderedDict([('b', 1), ('a', 2)])` vs `{'b': 1, 'a': 2}`)

For Python 2.7-3.6, `OrderedDict` is part of `collections` module, which simplifies this import.


jswhit:
Looks good. thanks!

PR #956, opened 2019-07-10
Remove underline from variable name in Dataset string repr

mwtoews:
While preparing #950, variable names were underlined using cleaver ANSI escape sequences. However, this was only done for Dataset, and not other classes. It is generally preferred to have `__repr__` return common strings (e.g. ascii), so I'm suggesting these to be removed. Not all interfaces support ANSI escape codes, so a odd-looking  `ESC[4mcmplx_varESC[0m` will be returned instead of `cmplx_var`.

However, happy to discuss further.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,9 @@ build/
 dist/
 *.egg-info/
 netCDF4/_netCDF4.c
+netCDF4/*.so
 include/constants.pyx
 netcdftime/_netcdftime.c
+venv/
+.eggs/
+.idea/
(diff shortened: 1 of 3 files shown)


jswhit:
I honestly don't remember why those are in there - so it's fine with me if they are removed.  Could you add a Changelog entry for this and your other PRs?

mwtoews:
I've added a second commit here (for convenience) which covers all of the recent issues/PRs that I've addressed over the past week.

Also, some items were added the `.gitignore` to hide files created from running `python setup.py build_ext -i` (which compiles the .so files in netCDF4/), and other directories created by PyCharm and virtualenv.

jswhit:
Thanks! merging now...

PR #960, opened 2019-07-17
Remove newlines most __repr__ methods

mwtoews:
This was another thing that I noted with #950, is that there are extra blank lines after printing various netCDF4 objects. Most Python `__repr__` methods generally don't end with `\n`, so this PR effectively removes the trailing `\n` that adds those extra blank lines.

Some of the code around the edited sections were re-organized to help readability and reduce repetition. But these string representations should otherwise remain the same with this PR.

Also update docstrings that show Python 3.7 output with `dict` (rather than `OrderedDict`); see #955


jswhit:
Looks good, thanks @mwtoews !

PR #1024, opened 2020-07-03
Replaces use of numpy's tostring() with tobytes()

### LINKED ISSUE #1023: Method numpy.ndarray.tostring() is deprecated in favor of tobytes() in 1.9
This deprecated (see at  accepted pull request https://github.com/numpy/numpy/pull/15867) method is used 6 times in the file  netCDF4/_netCDF4.pyx and may be somewhere else so each call to this method causes output of "DeprecationWarning: tostring() is deprecated. Use tobytes() instead." to stderrr.

jameshiebert:
numpy has deprecated the use of array.array.tostring() in favor of
tobytes
https://github.com/numpy/numpy/pull/15867
This patch replaces all (6) such uses in this package.

Fixes #1023 

I haven't done extensive testing, but the test suite runs fine on my dev machine:

```
(venv) james@basalt:~/code/git/netcdf4-python/test$ python run_all.py 
not running tst_unicode.py ...
not running tst_create_mem.py ...
not running tst_cdf5.py ...

netcdf4-python version: 1.5.4
HDF5 lib version:       1.10.0
netcdf lib version:     4.6.0
numpy version           1.19.0
cython version          0.29.20
......../home/james/code/git/netcdf4-python/test/tst_types.py:92: UserWarning: WARNING: missing_value not used since it
cannot be safely cast to variable data type
  assert_array_equal(v3[:],-1*np.ones(n2dim,v3.dtype))
..................................................................................foo_bar
.
----------------------------------------------------------------------
Ran 91 tests in 4.479s

OK
```

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -9,6 +9,8 @@
  * check size on valid_range instead of using len (issue #1013).
  * add `set_chunk_cache/get_chunk_cache` module functions to reset the
    default chunk cache sizes before opening a Dataset (issue #1018).
+ * replace use of numpy's deprecated tostring() method with tobytes()
+   (issue #1023).
 
  version 1.5.3 (tag v1.5.3rel)
 ==============================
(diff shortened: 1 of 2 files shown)


jswhit:
Looks good thanks - could you just add a Changelog entry?

PR #1028, opened 2020-07-21
prepare for v1.5.4 release

jswhit:


jswhit:
The builds on https://github.com/macpython/netcdf4-python are broken, since the server we have been using to stage the wheels no longer works (https://github.com/MacPython/netcdf4-python-wheels/issues/8).  

This means the 1.5.4 release is delayed.

PR #1029, opened 2020-07-22
also search include dir parents for hdf5

sroet:
While trying to do a development install inside conda like this:
```
conda create -n test python=3
conda activate test
conda install netcdf4 --only-deps
pip install -e .
```
pip would find the installed libraries of `libnetcdf` inside the conda environment, but not for `hdf5`.

This PR makes it so that `inc_dirs` are also searched to look for hdf5, this is done before the standard dirs to make sure it aligns with where `libnetcdf` is found.

This is only done, if none if the environment variables is set.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -330,7 +330,9 @@ def _populate_hdf5_info(dirstosearch, inc_dirs, libs, lib_dirs):
         if HDF5_incdir is None and HDF5_dir is None:
             sys.stdout.write("""
     HDF5_DIR environment variable not set, checking some standard locations ..\n""")
-            for direc in dirstosearch:
+            # Add parents of include dir to dirstosearch and look there first
+            parentdirs = [os.path.join(direc, os.pardir) for direc in inc_dirs]
+            for direc in parentdirs + dirstosearch:
                 sys.stdout.write('checking %s ...\n' % direc)
                 hdf5_version = check_hdf5version(os.path.join(direc, 'include'))
                 if hdf5_version is None:


setup.py
```diff
@@ -330,7 +330,9 @@ def _populate_hdf5_info(dirstosearch, inc_dirs, libs, lib_dirs):
         if HDF5_incdir is None and HDF5_dir is None:
             sys.stdout.write("""
     HDF5_DIR environment variable not set, checking some standard locations ..\n""")
-            for direc in dirstosearch:
+            # Add parents of include dir to dirstosearch and look there first
+            parentdirs = [os.path.join(direc, os.pardir) for direc in inc_dirs]
+            for direc in parentdirs + dirstosearch:
                 sys.stdout.write('checking %s ...\n' % direc)
                 hdf5_version = check_hdf5version(os.path.join(direc, 'include'))
```
sroet, line 337: We use the parent dirs to prevent extra logic on this check

sroet:
[this line already identifies the right include dir](https://github.com/Unidata/netcdf4-python/blob/c3413d2f00f65e13f39206e1bb0415da6f336166/setup.py#L379)

jswhit:
I don't understand the rationale for searching on level up from inc_dirs first.  Can you explain a little more what the problem is that this is trying to solve?

sroet:
What I am trying to solve:
this should work without any extra environment variables
```
conda create -n test python=3
conda activate test
conda install netcdf4 --only-deps
pip install -e .
```

# Why should it work?
in the `if not retcode` path of `setup.py`
the lines:
```python
    dep = subprocess.Popen([ncconfig, '--cflags'],                              
                           stdout=subprocess.PIPE).communicate()[0]             
    inc_dirs = [str(i[2:].decode()) for i in dep.split() if                     
                i[0:2].decode() == '-I']                                        
````
actually find the right include dir for libnetcdf, and it also contains the required files for HDF5
(adding a print statement and running `pip` with `-vvv` prints: `['/home/sroet/miniconda3/envs/test/include']`)
<details><summary> `ls` of that include dir (click me)</summary>
<p>

```
atom.h          H5DataType.h         H5MMpublic.h           jerror.h             tclUnixPort.h
bitvect.h       H5DcreatProp.h       H5Object.h             jmorecfg.h           tdbcDecls.h
bzlib.h         h5d.mod              H5OcreatProp.h         jpeglib.h            tdbc.h
cdeflate.h      H5DOpublic.h         h5o.mod                kadm5                tdbcInt.h
cnbit.h         H5Dpublic.h          H5Opublic.h            kdb.h                termcap.h
cnone.h         h5ds.mod             H5overflow.h           krad.h               term_entry.h
com_err.h       H5DSpublic.h         H5PacketTable.h        krb5                 term.h
crle.h          H5DxferProp.h        H5PLextern.h           krb5.h               tic.h
cskphuff.h      h5e.mod              H5PLpublic.h           ks_names.h           tk3d.h
cszip.h         H5EnumType.h         h5p.mod                libssh2.h            tkArray.h
curl            H5Epubgen.h          H5Ppublic.h            libssh2_publickey.h  tkBusy.h
cursesapp.h     H5Epublic.h          H5PredType.h           libssh2_sftp.h       tkButton.h
cursesf.h       H5Exception.h        H5PropList.h           linklist.h           tkCanvas.h
curses.h        H5f90i_gen.h         H5PTpublic.h           local_nc.h           tkColor.h
cursesm.h       H5f90i.h             H5pubconf.h            lzma                 tkDecls.h
cursesp.h       H5FaccProp.h         H5public.h             lzma.h               tkDList.h
cursesw.h       H5FcreatProp.h       h5r.mod                menu.h               tkEntry.h
cursslk.h       H5FDcore.h           H5Rpublic.h            mfan.h               tkFileFilter.h
default.h       H5FDdirect.h         h5s.mod                mfdatainfo.h         tkFont.h
dfan.h          H5FDfamily.h         H5Spublic.h            mfgr.h               tk.h
dfgr.h          H5FDhdfs.h           H5StrType.h            mfhdf.h              tkImgPhoto.h
df.h            H5FDlog.h            h5tb_const.mod         mfhdfi.h             tkIntDecls.h
dfi.h           H5FDmpi.h            h5tb.mod               mstdio.h             tkInt.h
dfrig.h         H5FDmpio.h           H5TBpublic.h           mysqlStubs.h         tkIntPlatDecls.h
dfsd.h          H5FDmulti.h          h5t.mod                nc_tparm.h           tkIntXlibDecls.h
dfstubs.h       H5FDpublic.h         H5Tpublic.h            ncurses              tkMacOSXConstants.h
dfufp2i.h       H5FDros3.h           H5VarLenType.h         ncurses_dll.h        tkMacOSXCursors.h
dynarray.h      H5FDsec2.h           H5version.h            ncurses.h            tkMacOSXDebug.h
editline        H5FDstdio.h          h5z.mod                ncursesw             tkMacOSXDefault.h
eti.h           H5FDwindows.h        H5Zpublic.h            netcdf_aux.h         tkMacOSXEvent.h
etip.h          H5File.h             hbitio.h               netcdf_dispatch.h    tkMacOSXFont.h
fakemysql.h     H5FloatType.h        hchunks.h              netcdf_filter.h      tkMacOSX.h
fakepq.h        h5f.mod              hcomp.h                netcdf.h             tkMacOSXInt.h
fakesql.h       h5fortkit.mod        hcompi.h               netcdf_mem.h         tkMacOSXPort.h
ffi.h           h5fortran_types.mod  hconv.h                netcdf_meta.h        tkMacOSXPrivate.h
ffitarget.h     H5Fpublic.h          hdatainfo.h            odbcStubs.h          tkMacOSXWm.h
form.h          h5_gen.mod           hdf2netcdf.h           openssl              tkMacOSXXCursors.h
gssapi          h5global.mod         hdf4_netcdf.h          panel.h              tkMenubutton.h
gssapi.h        h5g.mod              hdf5.h                 pqStubs.h            tkMenu.h
gssrpc          H5Gpublic.h          hdf5_hl.h              profile.h            tkPlatDecls.h
H4api_adpt.h    H5Group.h            hdf5.mod               python3.8            tkPort.h
h4config.h      H5IdComponent.h      hdf.h                  readline             tkScale.h
H5AbstractDs.h  h5im.mod             hdfi.h                 sqlite3ext.h         tkScrollbar.h
H5ACpublic.h    h5i.mod              herr.h                 sqlite3.h            tkSelect.h
h5a.mod         H5IMpublic.h         hfile.h                tbbt.h               tkText.h
H5api_adpt.h    H5Include.h          histedit.h             tclDecls.h           tkUndo.h
H5Apublic.h     H5IntType.h          hkit.h                 tcl.h                tkUnixDefault.h
H5ArrayType.h   H5Ipublic.h          hlimits.h              tclIntDecls.h        tkUnixInt.h
H5AtomType.h    H5LaccProp.h         hntdefs.h              tclInt.h             tkUnixPort.h
H5Attribute.h   H5LcreatProp.h       hproto.h               tclIntPlatDecls.h    unctrl.h
H5Classes.h     H5LDpublic.h         htags.h                tclOODecls.h         vattr.h
H5CommonFG.h    h5lib.mod            itcl2TclOO.h           tclOO.h              verto.h
H5CompType.h    H5Library.h          itclDecls.h            tclOOIntDecls.h      verto-module.h
H5CppDoc.h      h5l.mod              itcl.h                 tclOOInt.h           vg.h
H5Cpp.h         H5Location.h         itclIntDecls.h         tclPlatDecls.h       zconf.h
H5Cpublic.h     H5Lpublic.h          itclInt.h              tclPort.h            zlib.h
H5DaccProp.h    h5lt_const.mod       itclMigrate2TclCore.h  tclThread.h
H5DataSet.h     h5lt.mod             itclTclIntStubsFcn.h   tclTomMathDecls.h
H5DataSpace.h   H5LTpublic.h         jconfig.h              tclTomMath.h
```

</p>
</details>

# Why doesn't it work?

The next thing `setup.py` tries to find is hfd5, calling `_populate_hdf5_info` even with this `inc_dirs`
by default this goes down to  `else:-> if HDF5_incdir is None and HDF5_dir is None:` path 
It then tries to find HDF5 in the default directories, without first checking if `inc_dirs` already contains HDF5.
 
This leads to two possible bad outcomes (on master):
1. It does not find HDF5 in the default locations: it errors out with ` ValueError: did not find HDF5 headers` even though they are already part of the include_dirs
2. It does find HDF5 in the default location: This now leads to race conditions, because now 2 include dirs with valid HDF5 headers are given to the compiler, and by order it will look in the include that came from libnetcdf first, linking to that (even though netcdf4-python registers the HDF5 version it finds in the location it found in one of the default dirs). This leads to undefined behavior at runtime (depending on the order of installation) as described [in this comment](https://github.com/Unidata/netcdf4-python/pull/1026#discussion_r457647734)

Now, that comment indicates that this breaks even with an explicit HDF5_DIR during built time, and **this PR does not fix that** (I can submit another PR for that as well, if interested). 

This PR just makes it more sensible in the default case (no extra env variables given)

# Why do we go for the parent dir and search it first?
The reason why we go from the parent of the inc_dirs is because we want the line:
`hdf5_version = check_hdf5version(os.path.join(direc, 'include'))` to search `/home/sroet/miniconda3/envs/test/include` and not  `'/home/sroet/miniconda3/envs/test/include'/include`.

 The reason why we want to search it **first** and in the order of `inc_dirs` is because the compiler will also search for the include headers in that order, otherwise leading to the bad behavior point 2 anyway.

Please let me know if anything is still unclear.

jswhit:
OK - thanks for the detailed explanation of the problem.  Would an alternate solution be to not search any further for HDF5 if nc-config provides the necessary info?

sroet:
> Would an alternate solution be to not search any further for HDF5 if nc-config provides the necessary info?

I have not read all the code called throughout setup.py (I don't know what exactly nc-config is doing, just know that it returns the right dir for libnetcdf in my usecase), but we should at least check if nc-config is giving the hdf5 headers as well.

The other issue is that we don't want to override any expected behavior from having HDF5_DIR set, which seems like it might happen with what you are proposing, if I understand correctly. 

Now, this PR fixes my issue **only for the case where no environment variables are set**. 

More discussion needs to happen for the case where they **are** set, in order to make sure we don't run into the same undefined behavior, where headers are given more than once to the compiler. If you want, I can open a draft PR to discuss that further, but that is out of my intended scope for this PR.

jswhit:
PR #1032 checks the inc_dirs returned by nc-config for hdf5, and if it is found _populate_hdf5_info is not called.

@sroet - could you check to see if this resolves your original problem?  If not, we can perhaps combine these two PRs.

Right now, here's how setup.py manages the different ways of specifying directory locations

1) if USE_NCCONFIG is set in the environment, or use_ncconfig=True in setup.cfg, nc-config will be used and everything else will be ignored.

2) if USE_NCCONFIG is not set, env vars will be checked.  The info in setup.cfg will override this if USE_SETUPCFG is set and setup.cfg exists.

However, _populate_hdf5_info will still be called if nc-config is not used, and can potential over-ride values that are set by env vars or in setup.cfg.  This should still be fixed.  pkg-config is also used in some cases.  I don't understand how this works, and I suspect it probably is broken and should be removed.

I propose the following:

Keep the priority given in #1 and #2 above, but get rid of pkg-config.  Only call _populate_hdf5_info as a last resort if no other methods for finding hdf5 and netcdf are specified or nc-config doesn't work.

sroet:
#1032 solves the issue I was trying to solve with this PR :) (the default install now works as expected)

I will close this PR and move follow up discussion there.

Issue #1053, opened 2020-12-16
Migrate from travis CI to github actions


jswhit:
given the direction that travis is taking, and the fact that CI builds are taking longer and longer, we should migrate to github actions.   The [wheel-building repo](https://github.com/MacPython/netcdf4-python-wheels) also needs to move to github actions.

ocefpaf:
I have some experience with this. Do you want me to give it a go or do you have plans on doing it?

jswhit:
I have no plans to work on this - so I'm thankful you have stepped.

jswhit:
For the wheel-building repo, [dipy-wheels](https://github.com/MacPython/dipy-wheels) has already done this and we should be able to copy what they've done.

jswhit:
migration completed

PR #1054, opened 2020-12-17
Move AppVeyor tests to GitHub Actions

ocefpaf:
Partially addresses #1053

### DIFF
--- a/.github/workflows/miniconda-win.yml
+++ b/.github/workflows/miniconda-win.yml
@@ -1,0 +1,38 @@
+name: windows
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+
+jobs:
+  run:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.9]
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setup Conda
+      uses: s-weigand/setup-conda@v1
+      with:
+        activate-conda: false
+        conda-channels: conda-forge
+
+    - name: Python ${{ matrix.python-version }}
+      shell: bash -l {0}
+      run: |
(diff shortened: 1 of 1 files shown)


ocefpaf:
@jswhit I'm not aware of a way to activate GitHub Actions without merging one first. I'm happy to follow up on this PR and fix it later but to activate it we would need to merge it first.

.github/workflows/miniconda-win.yml
```diff
@@ -0,0 +1,38 @@
...
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.9]
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setup Conda
+      uses: s-weigand/setup-conda@v1
```
dopplershift, line 19: Any particular reason you've opted for this one and not [conda-icubator/setup-miniconda](https://github.com/conda-incubator/setup-miniconda)?

ocefpaf: It is less bloated and does less, just activate the pre-installed conda for you. Which is faster. I also had multiple issues with all the yaml options in the other one. I'd rather have the same commands on the config as those that I type on a terminal to have a 1-1 comparison if I need to debug. (With that said, I do break that rule here adding the conda-forge channel with a yaml config option :smile:)

dopplershift: Makes sense.

jswhit:
> @jswhit I'm not aware of a way to activate GitHub Actions without merging one first. I'm happy to follow up on this PR and fix it later but to activate it we would need to merge it first.

Thanks @ocefpaf - let me know when you are ready to merge and I will then merge and activate.

ocefpaf:
If you can let's merge this now and I'll work on it tomorrow with once the status is showing on PR.

PR #1057, opened 2020-12-21
Let's see if we can get this one to work

ocefpaf:
- running all the tests (removed set NO_NET=1)
- removed AppVeyor config file


ocefpaf:
@jswhit this is ready for review. I could not add Python 3.9 b/c that fails with older numpy 1.17. Based on https://numpy.org/neps/nep-0029-deprecation_policy.html I believe it is more important to test numpy 1.17 than Python 3.9, which is not that much different from 3.8.

jswhit:
building on what you have done, I think I've got everything (including the MPI tests) working in PR #1061.  If you have time, please give it a review.

PR #1061, opened 2020-12-24
run on windows, macosx and ubuntu

jswhit:


.github/workflows/miniconda.yml
```diff
@@ -24,8 +28,9 @@ jobs:
     - name: Python ${{ matrix.python-version }}
       shell: bash -l {0}
       run: |
-        conda create --name TEST python=${{ matrix.python-version }} "numpy=1.17" cython pip pytest hdf5 libnetcdf cftime --strict-channel-priority
+        conda create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest hdf5 libnetcdf cftime --strict-channel-priority
         source activate TEST
+        export PATH="${CONDA_PREFIX}/bin:${CONDA_PREFIX}/Library/bin:$PATH" # so setup.py finds nc-config
```
ocefpaf, line 33: Ah! The GHA was not adding this to the path. This was driving me nuts!

.github/workflows/build.yml
```diff
@@ -0,0 +1,103 @@
...
+    - name: Install python dependencies via pip
+      run: |
+        python -m pip install numpy
+        python -m pip install cython
+        python -m pip install cftime
+        python -m pip install pytest
+        python -m pip install check-manifest
+        python -m pip install twine
+        python -m pip install wheel
+        python -m pip install mpi4py
```
ocefpaf, line 69: I don't understand the new pip solver but maybe installing all of these in a single go can be faster than making it solve the current env for a future change in every new installation.

ocefpaf (approved):
Awesome! Sorry I dropped the ball on #1057 but I'm glad it was a little helpful to get this one done.

.github/workflows/build.yml
jswhit: Done

jswhit:
Thanks @ocefpaf  - most of this is just copied from PR #1057, so I wouldn't say you dropped the ball. I just carried it the last few meters.

Issue #1065, opened 2021-01-06
numpy 1.20 deprecates np.bool etc.


mathause:
The upcoming numpy 1.20 release will deprecate `np.bool` and `np.float` etc., see https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

You have some usages of this in your codebase that may throw `DeprecationWarning`s, we see that in the upstream-dev tests of xarray.  

E.g.:

https://github.com/Unidata/netcdf4-python/blob/bcbe3a0c62e6aa012b7afe7c9b2cfc9b2abc5424/src/netCDF4/_netCDF4.pyx#L4559

https://github.com/Unidata/netcdf4-python/blob/bcbe3a0c62e6aa012b7afe7c9b2cfc9b2abc5424/src/netCDF4/_netCDF4.pyx#L4930

I think the idea is to replace these with `bool` or `np.bool_` (I am never entirely sure what the difference is).

jswhit:
closed by PR #1066

PR #1066, opened 2021-01-13
deal with deprecated numpy.bool, float

jswhit:
closes issue #1065


mattip:
@eric-wieser what is wrong?

eric-wieser:
Find and replace was used without word boundaries and converted `numpy. float64` into `numpy.float_64` which does not exist

Issue #1073, opened 2021-01-26
License issue since cftime split


molinav:
Until version 1.4, the `netCDF4-python` library was released as a unique package including the `cftime` module, and the library was distributed with a license that resembles the Historical Permission Notice and Disclaimer (HPND) license (see https://opensource.org/licenses/HPND), which is very similar to the MIT license that was set recently in the upstream code.

Since version 1.4, the specific code for `cftime` was split into a separate repository (https://github.com/Unidata/cftime), and later `cftime` added a small portion of GPL code to handle the leap years in version 1.0.2. The inclusion of this code implies that `cftime` changed its license from HPND to GPL when this code snippet was added (as `cftime` would be considered a derivative work of this GPL snippet).

`netCDF4-python` is officially released currently using the MIT license, but since `netCDF4-python` has `cftime` as a dependency and `cftime` is released as GPL (as stated above), `netCDF4-python` would be considered a derivative work of `cftime` and therefore `netCDF4-python` can only have `cftime` as dependency if `netCDF4-python` is also released as GPL (note the difference with dynamic linking between GPL and LGPL).

I am not sure if the GPL license is intended for `netCDF4-python`, but I would assume it is not intended since `netCDF4-python` explicitly sets the license metadata as MIT in PyPI (https://pypi.org/project/netCDF4/1.5.5.1/). The current situation with the license may cause that people are downloading and using `netCDF4-python` thinking that the license is MIT but it actually inherits GPL from `cftime`, and this may be problematic for any project that is using `netCDF4-python` and that is not planned to be released as GPL.

At the moment, the only way to keep the old behaviour (HPND license/MIT-like license) is to:
* Keep `netCDF4-python` pinned in the requirements to `< 1.4`.
* Use a newer `netCDF4-python`, i.e. `>= 1.4`, but explicitly set the dependency `cftime >= 1.0, < 1.0.2` in the requirements.

Possible solutions to the current license inconsistency would be:
* Update the `netCDF4-python` license to indicate that the package is GPL (probably not intended by the `netCDF4-python` developers).
* Request the `cftime` developers to replace the GPL snippet with equivalent code that has a license that is not propagated to derivative works.

jswhit:
cftime uses some code that is derived from  [calcalcs](http://meteora.ucsd.edu/~pierce/calcalcs/index.html) that is GPL. Basically the calcalcs C code was translated to cython.   I've asked the author of calcalcs if he is willing to change the license, but got no response.

[...]

PR #1074, opened 2021-01-27
make cftime an optional dependency

jswhit:
addresses issue #1073

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,7 +56,7 @@ jobs:
     - name: Install python dependencies via pip
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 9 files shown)


ocefpaf:
I'm not a lawyer but I'm not sure this is needed. I believe that only linking would be affected by GPL, not direct dependencies.

jswhit:
If you google this, you'll find it's somewhat controversial.  The consensus seems to be that importing a python module is equivalent (from a legal perspective) to linking a dynamic library in a C program.

ocefpaf:
> If you google this, you'll find it's somewhat controversial. The consensus seems to be that importing a python module is equivalent (from a legal perspective) to linking a dynamic library in a C program.

If that is true an optional import would make this dual license and users should be warned. Licensing is hard :weary:

Issue #1078, opened 2021-02-04
Feature request: Support for import/export of CDL header


headtr1ck:
Would it be possible to implement the functionality of _ncdump_ and _ncgen_ into the python module?

Something like `netCDF4.Dataset.from_cdl(file_path)` and `dataset.to_cdl(file_path, data=False)` could be possible.

We often have to create or manipulate complex datasets and the cdl interface of the ncgen utility is quite useful to keep the same netCDF structure between different projects or languages, while being flexible enough to edit things before creating the file. However, in python the only possibility known to me is to call _ncgen_ in a subprocess and then open the netcdf file for editing.

jswhit:
https://github.com/rockdoc/cdlparser/wiki

jswhit:
also cf-python can apparently handle CDL

https://ncas-cms.github.io/cf-python/

davidhassell:
cf-python does indeed handle CDL, including abbreviated output, as you might get with `ncdump -h` or `ncdump -c`. It uses ncgen and subprocess internally, rather than trying to parse it natively:

https://ncas-cms.github.io/cf-python/function/cf.read.html

"... It [the CDL file] is converted to a temporary netCDF4 file using the external ncgen command, and the temporary file persists until the end of the Python session, at which time it is automatically deleted. The CDL file may omit data array values (as would be the case, for example, if the file was created with the -h or -c option to ncdump), in which case the the relevant constructs in memory will be created with data with all missing values."

jswhit:
rockdoc doesn't seem to be actively maintained anymore, and incorporated it into netcdf4-python would add a dependency on ply, which is a non-starter.

We could relatively easily incorporate the functionality from [cfdm](https://github.com/NCAS-CMS/cfdm/blob/master/cfdm/read_write/netcdf/netcdfread.py) to open a cdl file using ncgen spawned by sub-process.  Outputting cdl could be done similarly.  Or, this could be left to high-level external modules, like cf-python.    I welcome any opinions.

jswhit:
quick and dirty subprocess implementation in PR #1081.  CDL file import is via `netCDF4.dataset_fromcdl` function.  CDL export is via `Dataset.ncdump` method.

davidhassell:
Thanks for adding this, @jswhit  - I shall be replacing the `cfdm` read-CDL implementation with the new netCDF4 capability.

jswhit:
closed by PR #1081

PR #1081, opened 2021-02-08
cdl support via subprocess call to ncdump/ncgen

jswhit:
addresses issue #1078


headtr1ck:
Any reason why fromcdl was moved to dataset_fromcdl? I think class methods of the type cls.from_xxx(...) are quite common.
Also: thanks for the PR!

jswhit:
> Any reason why fromcdl was moved to dataset_fromcdl? I think class methods of the type cls.from_xxx(...) are quite common.
> Also: thanks for the PR!

I could make it a static method - but the docstring formatter we use (pdoc) doesn't extract docstrings from static methods.  The functionality is the same whether it is a module function or a static method of Dataset.

jswhit:
> > Any reason why fromcdl was moved to dataset_fromcdl? I think class methods of the type cls.from_xxx(...) are quite common.
> > Also: thanks for the PR!
> 
> I could make it a static method - but the docstring formatter we use (pdoc) doesn't extract docstrings from static methods. The functionality is the same whether it is a module function or a static method of Dataset.

I figured out how to get pdoc to extract static method docstrings, so I moved fromcdl back to a static method.

PR #1101, opened 2021-04-16
update tst_atts.py to use tocdl method instread of calling ncdump dir…

jswhit:
…ectly


jswhit:
failing tests on windows fixed by using file name in pwd, instead of NamedTemporaryFile

PR #1118, opened 2021-06-15
MAINT: remove a few Python2 features, mostly u'string' -> 'string'

mwtoews:
A few bits Python 2 remained, which are mostly cleaned up in this PR, and is mostly changing `u'string'` to `'string'`.

Removed `tst_unicode3.py` which was nearly identical to `tst_unicode.py`.


jswhit:
You don't need to change the html files, they are autogenerated from the docstrings.

Otherwise looks good.  Thanks for these updates.

mwtoews:
index.html is restored, so this PR is now ready to go.

It turns out the source for the former edit was cftime [here](https://github.com/Unidata/cftime/blob/6e2d752a4097dfb90623c3c97d4c57348c72a3a8/src/cftime/_cftime.pyx#L422-L431), but it doesn't seem to have `u'standard'`, so I'm not sure how this is integrated to index.html

PR #1134, opened 2021-10-15
Use assertRaisesRegex instead of assertRaisesRegexp for Python 3.11 compatibility.

tirkarthi:
The deprecated aliases were removed in Python 3.11 in python/cpython#28268 .


tirkarthi:
I am not really comfortable giving out my phone number for CLA since this is a tiny patch. Feel free to close or merge the PR as needed.

PR #1140, opened 2021-11-12
add support for quantization/bit-grooming in netcdf-c 4.8.2

jswhit:
WIP - using "significant_digits" kwarg in createVariable.

Question: what do to about old "least_signficant_digit" kwarg?  Retain for compatibility?  Issue warning if both specified?

### DIFF
--- a/.github/workflows/build_master.yml
+++ b/.github/workflows/build_master.yml
@@ -55,6 +55,9 @@ jobs:
       run: |
         export PATH=${NETCDF_DIR}/bin:${PATH} 
         python checkversion.py
+        cd examples
+        python bench_compress4.py
+        cd ..
         # serial
         cd test
         python run_all.py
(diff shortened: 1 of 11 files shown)


jswhit:
For now I've retained both the old 'DSD' quantization (via `least_significant_digit`) and the new NSD quantization in netcdf-c 4.8.2 via `significant_digits`.  If https://github.com/Unidata/netcdf-c/pull/2130 is merged, a way to specify the quantization algorithm will also be needed.

jswhit:
Unidata/netcdf-c#2130  enabled an alternate method for quantization.  This PR has been updated to use that alternative method if the `significant_digits` kwarg is negative.  According to the discussion in Unidata/netcdf-c#2130, there should never be more than two options for quantization (the `NC_QUANTIZE_GRANULARBG` option should point to the current state-of-the art algorithm), this should be sufficient.

PR #1146, opened 2022-01-09
change behavior of mode='a' to be like python open (issue #1144)

jswhit:


zerothi:
I would love this. :)
Thanks!

src/netCDF4/_netCDF4.pyx
```diff
@@ -2246,7 +2255,7 @@ strings.
                     ierr = nc_open(path, NC_NOWRITE | NC_SHARE, &grpid)
                 else:
                     ierr = nc_open(path, NC_NOWRITE, &grpid)
-        elif mode == 'r+' or mode == 'a':
+        elif mode == 'r+' or mode == 'a' and os.path.exists(filename):
```
zerothi, line 2258: `and` and `or` does not have the same precedence, this is read as `mode == r+ or (mode == a and os.path.exists(filename)`, didn't you mean: `mode in ('r+', 'a') and os.path.exists(filename)`? Same problem further down.

zerothi, line 2258: In any case, putting parenthesis might be more clear about intent

jswhit2: good catch - will fix

Issue #1145, opened 2022-01-07
createVariable does not accept `Dimension` objects as dimension specifications


zerothi:
It would be nice if the following could be done, for completeness sake:
```python
import netCDF4 as nc

out = nc.Dataset("hello.nc", 'w')
dim = out.createDimension("size", 2)
foo = out.createVariable("foo", 'f4', ('size',))
bar = out.createVariable("bar", 'f4', (dim,))
```
The documentation is coherent with the implementation, but I think the above should be allowed.

jswhit:
Not obvious to me that this is necessary, but if someone's willing to implement this I'll entertain a PR.

jswhit:
This actually turned out to be pretty straightforward as well, so I created a draft PR #1147

jswhit:
PR #1147 merged, closing now...

PR #1147, opened 2022-01-15
allow createDimension to accept Dimension instances (issue #1145)

jswhit:


src/netCDF4/_netCDF4.pyx
```diff
@@ -2757,6 +2758,17 @@ is the number of variable dimensions."""
             group = self
         else:
             group = self.createGroup(dirname)
+        # if dimensions is a single string or Dimension instance, 
+        # convert to a tuple.
+        # This prevents a common error that occurs when
+        # dimensions = 'lat' instead of ('lat',)
+        if type(dimensions) == str or type(dimensions) == bytes or\
```
zerothi, line 2765: why not just `isinstance(dimensions, (str, bytes, Dimension))`? type forces the type but not sub-classes?

src/netCDF4/_netCDF4.pyx
```diff
@@ -2653,8 +2653,9 @@ length greater than one are aliases for `str`.
 Data from netCDF variables is presented to python as numpy arrays with
 the corresponding data type.
 
-`dimensions` must be a tuple containing dimension names (strings) that
-have been defined previously using `Dataset.createDimension`. The default value
+`dimensions` must be a tuple containing `Dimension` instances or
```
zerothi, line 2656: perhaps to clarify that a mix is allowed, `a tuple containing mixed `Dimension`. Perhaps it's just me ;)

zerothi (commented):
Great! Thanks!

src/netCDF4/_netCDF4.pyx
jswhit: done

src/netCDF4/_netCDF4.pyx
jswhit: use and/or instead of or

PR #1149, opened 2022-01-20
deprecate zlib kwarg to createVariable, replace with compression kwarg

jswhit:
Only compression=None (or False) and compression='zlib' currently allowed.  This sets the stage for including other compression algorithms (such as 'zstandard') as they become available in netcdf-c (see https://github.com/Unidata/netcdf-c/issues/2173)


jswhit:
added new 'quantize_mode' kwarg to specify quantization method (instead of relying at sign of significant_digits, which assumes there will only ever be two methods).

Waiting on https://github.com/Unidata/netcdf-c/pull/2232 to be merged to add a third method `BitRound` (in additionn to `BitGroom` and `GranularBitRound`)

PR #1150, opened 2022-01-27
add /opt/homebrew to default search path

jswhit:


jswhit:
this is where [homebrew](http://brew.sh) installs stuff on Apple Silicon/ARM

PR #1156, opened 2022-03-05
Define MPI_Session for compatibility with current mpi4py main branch

francesco-ballarin:
As of 12 days ago, `mpi4py` has a new define `MPI_Session` which must be added to `include/mpi-compat.h` to prevent a `error: unknown type name ‘MPI_Session’` upon compiling `src/netCDF4/_netCDF4.c`. 

Also see the blame
https://github.com/mpi4py/mpi4py/blame/80c53a02c2f85c721c8c26b202a4ab0ed4ac02fc/src/mpi4py/include/mpi4py/mpi4py.h

### DIFF
--- a/.github/workflows/miniconda.yml
+++ b/.github/workflows/miniconda.yml
@@ -78,7 +78,7 @@ jobs:
         export PATH="${CONDA_PREFIX}/bin:${CONDA_PREFIX}/Library/bin:$PATH" 
         which mpirun
         mpirun --version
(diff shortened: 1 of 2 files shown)


francesco-ballarin:
@jswhit openmpi failure seems unrelated, but the second commit should fix it.

jswhit:
OK thanks @francesco-ballarin! Merging now...

Issue #1186, opened 2022-08-31
windows tests failing on master


jswhit:
cython compilation fails to find zlib

creating D:\a\netcdf4-python\netcdf4-python\build\lib.win-amd64-cpython-38
creating D:\a\netcdf4-python\netcdf4-python\build\lib.win-amd64-cpython-38\netCDF4
"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.33.31629\bin\HostX86\x64\link.exe" /nologo /INCREMENTAL:NO /LTCG /DLL /MANIFEST:EMBED,ID=2 /MANIFESTUAC:NO /LIBPATH:C:\Miniconda\envs\TEST\Library\lib /LIBPATH:C:\Miniconda\envs\TEST\Library\lib /LIBPATH:C:\Miniconda\envs\TEST\libs /LIBPATH:C:\Miniconda\envs\TEST /LIBPATH:C:\Miniconda\envs\TEST\PCbuild\amd64 "/LIBPATH:C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.33.31629\ATLMFC\lib\x64" "/LIBPATH:C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.33.31629\lib\x64" "/LIBPATH:C:\Program Files (x86)\Windows Kits\NETFXSDK\4.8\lib\um\x64" "/LIBPATH:C:\Program Files (x86)\Windows Kits\10\lib\10.0.22621.0\ucrt\x64" "/LIBPATH:C:\Program Files (x86)\Windows Kits\10\\lib\10.0.22621.0\\um\x64" hdf5_hl.lib hdf5.lib netcdf.lib zlib.lib /EXPORT:PyInit__netCDF4 build\temp.win-amd64-cpython-38\Release\src\netCDF4\_netCDF4.obj /OUT:build\lib.win-amd64-cpython-38\netCDF4\_netCDF4.cp38-win_amd64.pyd /IMPLIB:build\temp.win-amd64-cpython-38\Release\src\netCDF4\_netCDF4.cp38-win_amd64.lib
LINK : fatal error LNK1181: cannot open input file 'zlib.lib'
[...]

PR #1187, opened 2022-08-31
try micromamba

ocefpaf:
Debugging #1186

### DIFF
--- a/.github/workflows/miniconda.yml
+++ b/.github/workflows/miniconda.yml
@@ -67,7 +67,7 @@ jobs:
     - name: Python ${{ matrix.python-version }}
       shell: bash -l {0}
       run: |
-        micromamba create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest mpi4py hdf5=*=mpi* libnetcdf=*=mpi* cftime --channel conda-forge
+        micromamba create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest mpi4py hdf5=*=mpi* libnetcdf=*=mpi* cftime zlib --channel conda-forge
         micromamba activate TEST
         export PATH="${CONDA_PREFIX}/bin:${CONDA_PREFIX}/Library/bin:$PATH" # so setup.py finds nc-config
         pip install -e . --no-deps --force-reinstall
--- a/.github/workflows/miniconda.yml
+++ b/.github/workflows/miniconda.yml
@@ -18,29 +18,35 @@ jobs:
         exclude:
           - os: macos-latest  
             platform: x32
-    steps:
-    - uses: actions/checkout@v2
+      fail-fast: false
 
-    - name: Setup Conda
-      uses: s-weigand/setup-conda@v1
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Setup Micromamba
+      uses: mamba-org/provision-with-micromamba@v13
       with:
-        activate-conda: false
-        conda-channels: conda-forge
+        environment-file: false
 
     - name: Python ${{ matrix.python-version }}
       shell: bash -l {0}
       run: |
-        conda create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest hdf5 libnetcdf cftime --strict-channel-priority
-        source activate TEST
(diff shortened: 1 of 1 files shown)


ocefpaf:
@hmaarrfk I believe the changes in https://github.com/conda-forge/zlib-feedstock/pull/62 broke the Windows CIs here. Using the previous version or build number before that PR works. As the author of that PR do you mind taking a look here and help us figure this out?

@jswhit this PR adds a workaround for the zlib issue but we should unpin it when the problem is fixed upstream. I'll leave to you if we should merge this or not. There are other changes that are worth keeping regardless of decision on the zlib pin.

ocefpaf:
I believe the failure there is a network issue and should pass with restart.

hmaarrfk:
https://github.com/conda-forge/libnetcdf-feedstock/issues/146

hmaarrfk:
is the issue that libnetcdf fails, or that this build here fails?

ocefpaf:
> is the issue that libnetcdf fails, or that this build here fails?

The build here fails. The debug info is in https://github.com/Unidata/netcdf4-python/issues/1186#issue-1357408267

jswhit:
thanks for fixing this so quickly @ocefpaf !

hmaarrfk:
do you have a build that just add `zlib` instead of pinning it?

.github/workflows/miniconda.yml
```diff
@@ -18,29 +18,35 @@ jobs:
...
-        activate-conda: false
-        conda-channels: conda-forge
+        environment-file: false
 
     - name: Python ${{ matrix.python-version }}
       shell: bash -l {0}
       run: |
-        conda create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest hdf5 libnetcdf cftime --strict-channel-priority
-        source activate TEST
+        micromamba create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest hdf5 libnetcdf cftime "zlib=1.2.11" --channel conda-forge
```
hmaarrfk, line 34:
```suggestion
        micromamba create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest hdf5 libnetcdf cftime zlib --channel conda-forge
```
does this work?

.github/workflows/miniconda.yml
```diff
@@ -53,26 +59,30 @@ jobs:
...
-        activate-conda: false
-        conda-channels: conda-forge
+        environment-file: false
 
     - name: Python ${{ matrix.python-version }}
       shell: bash -l {0}
       run: |
-        conda create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest mpi4py hdf5=*=mpi* libnetcdf=*=mpi* cftime --strict-channel-priority
-        source activate TEST
+        micromamba create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest mpi4py hdf5=*=mpi* libnetcdf=*=mpi* cftime --channel conda-forge
```
hmaarrfk, line 70:
```suggestion
        micromamba create --name TEST python=${{ matrix.python-version }} numpy cython pip pytest mpi4py hdf5=*=mpi* libnetcdf=*=mpi* cftime zlib --channel conda-forge
```

.github/workflows/miniconda.yml
hmaarrfk: Deeper explanation; https://github.com/Unidata/netcdf4-python/issues/1186#issuecomment-1233339986

hmaarrfk:
feel free to squash.

ocefpaf:
> feel free to squash.

I'll leave that to @jswhit, not sure what he prefers. Thanks @hmaarrfk for the quick debug!

jswhit:
If the fix upstream will happen soon, I'd prefer to wait for that.  I'll leave this open in the meantime.  Which changes in this PR do we need even after the upstream fix?

ocefpaf:
> If the fix upstream will happen soon, I'd prefer to wait for that. I'll leave this open in the meantime. Which changes in this PR do we need even after the upstream fix?

Sorry. I should've said before. @hmaarrfk figured out the real issue and it is that we were depending on an indirect pull of `libzlib`. B/c netcdf4-python depends on zlib directly we have to make it an explicit dependency, no need for workarounds with previous versions.

This should be good to go as-is.

Issue #1192, opened 2022-09-15
Very frequent segfaults with the new `netCDF4=1.6.1`


valeriupredoi:
Heads up guys, we are seeing some very frequent segfaults when we have the new, hours-old, `netCDF4=1.6.1` in our environment. It's most probably due to it, since HDF5 has been at 1.12.2 for a while now - more than a month, and with `netCDF4=1.6.0` all works fine (and other packages staying at the same version and hash point). Apologies if this proves out to be due to a different package, but better safe than sorry in terms of a forewarning. Cheers muchly :beer:

ocefpaf:
Looks like you are using conda-forge's netcdf4. Maybe open an issue at https://github.com/conda-forge/netcdf4-feedstock instead.

PS: could you also test the wheels just to be sure they are OK?

valeriupredoi:
@ocefpaf good call, mate! Will do so, cheers :beer:

neutrinoceros:
We're also having issues on yt with the windows wheels for version 1.6.1. Namely, h5py is raising a warning at import. See https://github.com/yt-project/yt/issues/4128

trexfeathers:
Same with us - see SciTools/iris#4968. Sometimes manifests as segfaults, sometimes as crashed GHA workers (maybe segfault underneath).

@ocefpaf I've confirmed the same problems appear when installing from PyPI OR from conda-forge.

valeriupredoi:
> @ocefpaf I've confirmed the same problems appear when installing from PyPI OR from conda-forge.

Many thanks, I was about to test the PyPi version - cheers for testing, that saves me some lunch time :grin:

ocefpaf:
@trexfeathers what platforms are failing when you tested the PyPI wheels? I'm particularly interested in the Windows wheels for 1.6.1 b/c those are built in a different way now.

trexfeathers:
> @trexfeathers what platforms are failing when you tested the PyPI wheels? I'm particularly interested in the Windows wheels for 1.6.1 b/c those are built in a different way now.

GHA's `ubuntu-latest`. Despite several attempts we are yet to get Iris' test suite working on Windows.

neutrinoceros:
@ocefpaf to clarify, on yt we're testing with PyPI wheels for all three major platforms, and we're only seeing issues on windows.

valeriupredoi:
[...]

PR #1193, opened 2022-09-22
don't release GIL in Dataset.__init__

jswhit:
potential fix for issue  #1192

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,3 +1,7 @@
+ version 1.6.2 (not yet released)
+=================================
+ * do not release GIL for netcdf-c calls in Dataset.__init__ (to avoid segfaults in threaded code).
+
  version 1.6.1 (tag v1.6.1rel)
 ==============================
  * add Dataset methods has_<name>_filter (where <name>=zstd,blosc,bzip2,szip)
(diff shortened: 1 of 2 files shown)


dopplershift:
Personally, I think it's a mistake for users to rely on the GIL to make netcdf "threadsafe".

jswhit:
Understood @dopplershift. Perhaps the GIL may have been prevented some segfaults, but results still may have been incorrect.

ocefpaf:
What is the advantage to maintain this part of the code and let users fix it instead vs merging this to safe guard bad use? My impression is that most folks opening issue don't even know they are abusing the GIL there.

dopplershift:
Previously, the library was inconsistently releasing the GIL, as in: some methods were already releasing it and were thread unsafe. (I've personally encountered crashes as a result). So to truly solve the problem and eliminate all potential crashes due to Python-based multithreading, you'd need to not only revert #1181, but #370 as well, and make sure ALL calls hold the GIL.

Personally, I'd rather have the option for better performance for use cases that aren't trying to use multiple netCDF files, but doing other things concurrently in the interpreter. I do see that this change has been a bit disruptive, so I'm not going to fight for it. It's also not my library, so my opinion counts very little. :wink:

PR #1198, opened 2022-10-11
return empty array if boolean index array is all False

jswhit:
alternative to PR #1197


jswhit:
Behavior is consistent with numpy.  Also, indexing 1d variables with boolean index arrays that are all False already return empty arrays (but for variables with more than one dimension an exception is raised). With this PR the behavior for multi-dim variables is consistent with the current behavior for 1d variables (and slicing numpy arrays)

PR #1212, opened 2022-11-13
prepare for v1.6.2 release

jswhit:


ocefpaf:
> @ocefpaf could you create a windows wheel for the v1.6.2rel branch?

Done in https://github.com/ocefpaf/netcdf4-win-wheels/pull/7

Artifacts can be downloaded here: https://github.com/ocefpaf/netcdf4-win-wheels/actions/runs/3470462639


BTW, I don't mind doing this but we need to increase the bus factor on building Windows wheels.
I'd be happy to add you there as a maintainer on the original repo or transfer that it to Unidata. Whatever is easier to maintain.

jswhit:
If the long term solution is to build all the wheels with https://github.com/ocefpaf/netcdf4-win-wheels/pull/6, then I think moving it to the Unidata repo is the way to go.  @dopplershift what do you think - do you want to wait and try the cibuildwheel approach first?

ocefpaf:
> If the long term solution is to build all the wheels with [ocefpaf/netcdf4-win-wheels#6](https://github.com/ocefpaf/netcdf4-win-wheels/pull/6), then I think moving it to the Unidata repo is the way to go.

OK. Meanwhile would you and @dopplershift be OK to be added as owners of the netcdf4-win-wheels too? Just in case there are a few new releases until we sort this out. Updating there is relatively easy, just bump the version and checksum.

PR #1215, opened 2022-11-17
Drop Python 3.6, apply a few minor Python 3 styles

mwtoews:
Python 3.6 is no longer supported ([PEP 494](https://peps.python.org/pep-0494/)), and testing with that version was recently removed.

Provisionally, the minimum Python version is set to 3.7. However, it is recommended to push this to 3.8 to align with [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) for NumPy, which has been 3.8+ for nearly a year. Any thoughts on 3.7 or 3.8 as a minimum?

---
Other minor Python 3 styles:

- the `from __future__ import` statements were relevant for Python 2
- `IOError` is an alias of `OSError` since [PEP 3151](https://peps.python.org/pep-3151/)
- the default for `.encode()` is already utf-8, so does not need to be specified

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,7 @@ conda install -c conda-forge netCDF4
 * Clone GitHub repository (`git clone https://github.com/Unidata/netcdf4-python.git`)
 
 * Make sure [numpy](http://www.numpy.org/) and [Cython](http://cython.org/) are
-  installed and you have [Python](https://www.python.org) 3.6 or newer.
+  installed and you have [Python](https://www.python.org) 3.7 or newer.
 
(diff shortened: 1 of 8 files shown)


jswhit:
Looks reasonable to me, thanks

mwtoews:
I've just added another commit to remove unused code for Python<3.7 for OrderedDict.

One downside of keeping 3.7 is that more testing is required, and more binary wheels get uploaded to PyPI, which has a default limit of 10 GB, so this would spare ~18 MB per release. It's fine to keep 3.7 for now, but [PEP 537](https://peps.python.org/pep-0537/) has the last release in mid-2023.

PR #1216, opened 2022-11-20
Move static project metadata to pyproject.toml

mwtoews:
This PR moves the "[boring parts](https://peps.python.org/pep-0621/)" of this project's metadata to pyproject.toml, as supported by [setuptools>=61](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). This leaves setup.py to deal with the dynamic project build aspects.

This should be a close 1:1 conversion, with a few changes:

- The build system Cython version is pushed up to [0.29](https://pypi.org/project/Cython/0.29/), and setuptools must be [61](https://pypi.org/project/setuptools/61.0.0/) or later
- The minimum numpy version is removed, since it is probably irrelevant -- the previous limit of [1.9](https://pypi.org/project/numpy/1.9.0/) is very old
- The long_description / description text is wrapped (for readability), but also a content-type is provided to finally pass `twine check --strict dist/*` (it failed before without a content type). Also, the last line with a link to the repo is removed, as it is better provided in the project.urls section which is rendered nicely on PyPI
- url / download_url are replaced with a project.urls section, currently with documentation and repository links

Other considerations:

- This setup will not work with older setuptools versions, and requires python>=3.7
- It should not pose a problem with inclusion for debian/ubuntu, e.g. see https://packages.debian.org/testing/python3-setuptools which meets the minimum requirements


setup.py
```diff
@@ -726,37 +715,11 @@ def _populate_hdf5_info(dirstosearch, inc_dirs, libs, lib_dirs):
...
-                   "Operating System :: OS Independent"],
-      packages=find_namespace_packages(where="src"),
-      package_dir={'':'src'},
-      package_data={"netCDF4.plugins": ["lib__nc*"]},
-      ext_modules=ext_modules,
-      python_requires=">=3.7",
-      **setuptools_extra_kwargs)
+# See pyproject.toml for project metadata
+setup(
+    version=extract_version(netcdf4_src_pyx),
```
ocefpaf, line 720: I'm sorry I'm late here but I'd like to keep at least the name metadata in the `setup.py` for a while. The reason is b/c GitHub doesn't parse pyproject.toml yet and we lose the dependency graph info in https://github.com/Unidata/netcdf4-python/network/dependents without it.

https://github.com/community/community/discussions/6456

mwtoews: that seems like a GH bug. Besides discovering other connected projects, what is a dependency graph used for?

mwtoews: reading comments from the past two weeks, it appears that the Dependency Graph team are investigating a fix. I'd expect to see this resolved soon, particularly since the PEP is now supported by setuptools.

ocefpaf: It has been years since that was originally reported. I would not hold my breath. Also, re-adding just the name metadata in the setup.py is not that hard.

mwtoews: we could simply plop the relevant metadata to setup.py until it's resolved; see e.g. [flask](https://github.com/pallets/flask/blob/cc66213e579d6b35d9951c21b685d0078f373c44/setup.py#L3-L17). I'm in the middle of refactoring setup.py, so I can include that when ready.

ocefpaf: That is what I asked in my first message. Sorry if that was not clear.

mwtoews: yup, no probs I'll do that in that forthcoming PR, it just might be a few more days without a dependency graph.

PR #1219, opened 2022-11-24
fix #1218 for parsing non-existent values in netcdf_meta.h

### LINKED ISSUE #1218: Error building netcdf4 from source due to naive reading of /usr/include/netcdf_meta.h
My debian bullseye **/usr/include/netcdf_meta.h** file contains :

```
..
#define NC_HAS_HDF5      1 /*!< HDF5 support. */
#define NC_HAS_SZIP      /*!< szip support (HDF5 only) */
#define NC_HAS_SZIP_WRITE  /*!< szip write support (HDF5 only) */
#define NC_HAS_DAP2      1 /*!< DAP2 support. */
..
```

The naïve splitting into keys and values in `setup.py` meas that it interprets `/*!<` as the value for key `NC_HAS_SZIP` and `NC_HAS_SZIP_WRITE` which causes the build to fail.

This can be solved by either updating the values in the netcdf_meta.h header file or by improving the parsing of said file.

bluppfisk:
this just wraps any such reads in try/except blocks catching ValueError and leaving the default (False) assigned.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -114,23 +114,28 @@ def check_api(inc_dirs,netcdf_lib_version):
                 if line.startswith('#define NC_HAS_CDF5'):
                     try:
                         has_cdf5_format = bool(int(line.split()[2]))
-                    except ValueError: ...  # keep default False if value cannot be parsed
+                    except ValueError:
+                        pass  # keep default False if value cannot be parsed
                 if line.startswith('#define NC_HAS_PARALLEL'):
                     try:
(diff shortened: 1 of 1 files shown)


jswhit:
Thanks for catching this.  Seems like there needs to be a `pass` after each `except ValueError:` though.

bluppfisk:
hi @jswhit I agree that `pass` is better. I've updated the PR

bluppfisk:
tests are failing but I don't think we can ascribe that to the code change.

Issue #1220, opened 2022-12-01
PermissionError: [Errno 13] Permission denied: b'F:\\\xe5\x88\x9b\xe5\xbb\xba\xe6\x96\x87\xe4\xbb\xb6\\\\new.nc'


yqhhhh612:
To report a non-security related issue, please provide:

Software: pycharm educational 2021, netCDF4 1.4.0
Computer system and compile: windows10, python3.9.6

When run scripts as follow:
'''
from netCDF4 import Dataset
import numpy as np
savepath = r'F:\创建文件\\'
nc_data = Dataset(savepath+'new.nc','w',format='NETCDF4')
x = np.arange(0,10,1)
a = np.zeros((10,10))
nc_data.createDimension('x',10)
nc_data.createVariable('data',np.float32,('x','x'))
nc_data.variables['data'][:] = a
nc_data.close()
'''
It's seems that error reporting when file path is not in English. Although it is no longer wrong after being changed to English, I hope it will be more flexible

jswhit:
possibly related to https://github.com/Unidata/netcdf4-python/issues/997

jswhit:
PR #1221 fixes the printing of the filename in the error message.  @yqhhhh612 can you confirm that upgrading netcdf4-python fixes the problem?  (you can use `pip install --upgrade netCDF4`)

yqhhhh612:
I upgraded the netCDF4 version and the problem persists. The file path still cannot appear in Chinese or it will report an error.

PR #1221, opened 2022-12-05
Rename IOError alias to OSError, decode filename in message

mwtoews:
This PR does the following:

- Rename IOError and EnvironmentError aliases to OSError; see [PEP 3151](https://peps.python.org/pep-3151/) for background
- Decode filename for error message from _ensure_nc_success which will change messages e.g. difference with [#1220](https://github.com/Unidata/netcdf4-python/issues/1220):
  ```diff
  - FileNotFoundError: [Errno 2] No such file or directory: b'\xe5\x88\x9b\xe5\xbb\xba\xe6\x96\x87\xe4\xbb\xb6/file.nc'
  + FileNotFoundError: [Errno 2] No such file or directory: '创建文件/file.nc'
  ```
- <s>With test/tst_dap.py, check if URLs work before running tests, otherwise skip them</s>
  - This is primarily to resolve a failed CI test since https://remotetest.unidata.ucar.edu/thredds currently returns HTTP Status 404 – Not Found

### DIFF
--- a/src/netCDF4/_netCDF4.pyx
+++ b/src/netCDF4/_netCDF4.pyx
@@ -1712,7 +1712,7 @@ be raised in the next release."""
             # don't allow string array attributes in NETCDF3 files.
             if is_netcdf3 and N > 1:
                 msg='array string attributes can only be written with NETCDF4'
-                raise IOError(msg)
+                raise OSError(msg)
             if not value_arr.shape:
                 dats = _strencode(value_arr.item())
             else:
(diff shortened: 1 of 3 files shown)


jswhit:
you can skip the opendap tests by setting the env var NO_NET=1

dopplershift:
That server can be a little unstable. You might be better off updating the test to use some of our case study data on the production server (with which we try for much better uptime): https://thredds.ucar.edu/thredds/dodsC/casestudies/harvey/model/gfs_ana/GFS_Global_0p5deg_ana_20170829_1800.grib2

mwtoews:
Sure, we can let this test fail. I think the point of the first test (with `URL`) is to test non-secure HTTP. Is there another example DAP server that only uses HTTP? Or does anyone have a contact to repair the example URL?

jswhit:
PR #1223 updated the DAP test to use a different server.  You can merge in the update from master and remove that url check now.

mwtoews:
Ready for a final review.

Issue #1246, opened 2023-04-21
opendap access crash only with pip installation in ubuntu-latest SO


aragong:
Hi, I am suffering crashes in my github action tests only when I use netCDF4 opendap access (I tried several datasets in different urls/servers) installed with `pip install netCDF4` and `ubuntu-latest` machines.

I supposse that could be an xarray related error but seems to be a problem in the installation related with netCDF4. view issue in xarray repository: https://github.com/pydata/xarray/issues/7773

I also obtain this error if I use google colab platform that I suppose that use the same environment in ubunutu-latest SO.

![image](https://user-images.githubusercontent.com/48764870/233559427-12613dcf-dd44-4696-a671-9babb775332e.png)
![image](https://user-images.githubusercontent.com/48764870/233559526-95d22329-b06f-4fb5-b65e-b46e3ac4a14a.png)

I think that is related to this other opened issue... not sure! https://github.com/Unidata/netcdf4-python/issues/1179

Thank you in advance!

To report a non-security related issue, please provide:

* the version of the software with which you are encountering an issue
* environmental information (i.e. Operating System, compiler info, java version, python version, etc.)
* a description of the issue with the steps needed to reproduce it

If you have a general question about the software, please view our [Suggested Support Process](https://www.unidata.ucar.edu/support/#process).

jswhit:
Does ncdump work on that URL?  If so, then it's a python interface issue, if not it's a netcdf-c lib issue.

aragong:
I tried this in a [google colab script](https://colab.research.google.com/drive/1CDD13vxJJoz7_l7X9X6wJfLXk1JMeeIL?usp=sharing) 

1 - I intalled netcdf4 using pip --> OK
2 - ncdump is not installed in the ubuntu by default --> apt install netcdf-bin
3 - ncdump to the url works fine in terminal
4 - Dataset class in python raise an error
5 - Xarray also raises an error


![image](https://user-images.githubusercontent.com/48764870/233924682-1f11d05e-5b79-457c-bf8a-3dcb459f5e9c.png)
![image](https://user-images.githubusercontent.com/48764870/233924788-206b5af0-028b-4e68-8c9d-2721f350ac82.png)
![image](https://user-images.githubusercontent.com/48764870/233924910-704f6329-dc68-4d1a-93fc-41e6205d2971.png)
[...]

PR #1247, opened 2023-04-26
fix for https OpenDAP URLs with linux wheels (issue #1246)

jswhit:


jswhit:
the path to the SSL certificates is now set using certifi and the nc_rc_set netcdf-c function to fix the problem with https URLs in the linux wheels.

jswhit:
the fix also requires https://github.com/Unidata/netcdf-c/pull/2690

Issue #1119, opened 2021-06-15
Task: update examples notebook


mwtoews:
The Jupyter notebook examples are a bit old, and could use a refresh.

For instance `reading_netCDF.ipynb` has an outdated URL, which should be updated from:
```
URL = 'http://thredds.ucar.edu/thredds/dodsC/grib/NCEP/GFS/Global_0p5deg/GFS_Global_0p5deg_%04i%02i%02i_%02i%02i.grib2/GC'
```
to something like
```
URL = 'https://thredds.ucar.edu/thredds/catalog/grib/NCEP/GFS/Global_0p5deg/GFS_Global_0p5deg_%04i%02i%02i_%02i%02i.grib2/GC'
```
except there is a "HTTP Status 404" with the "/GC" part. I'm not sure the correct address, but clearly it's changed. Start [here](https://thredds.ucar.edu/thredds/catalog/grib/NCEP/GFS/Global_0p5deg/catalog.html) for the current address.

----

`writing_netCDF.ipynb` works nicely, so the steps are:

1. `jupyter notebook writing_netCDF.ipynb`
2. Cell > Run All
3. Save and Checkpoint

There are possibly other improvements to make, which is why I'm holding back.

jswhit:
pull request would be welcomed

PR #1260, opened 2023-05-25
fix URL

ocefpaf:
A friend just hit #1119 when trying out netCDF4.

There are other improvements that could be done here but for now all I did was to update the URL.
The changes are massive b/c the repo is storing the outputs. Maybe we could remove the outputs and add a binder link so folks can try these notebooks out.


jswhit2:
thanks - and agree about removing outputs and adding binder link.

PR #1266, opened 2023-07-09
Re-arrange pyproject.toml readme to be compatible with older parsers

mwtoews:
The content is the same, but re-arrange the items in `pyproject.toml` to work with older TOML parsers.

Closes #1265

Issue #1265, opened 2023-07-06
pip install netCDF4 for 1.6.3+ fails with TomlDecodeError using distro-default pip version


PhilMiller:
It looks like the last couple patch releases after 1.6.2 created a dependency on newer versions of pip that weren't previously required up through 1.6.2. That seems surprising and maybe inappropriate. The upgrade itself isn't burdensome, but the error is wholly inscrutable and doesn't at all steer toward that as a solution. It took me over an hour to guess to try the upgrade based on pip appearing in the stack trace, and a few Google searches didn't surface anything that suggested it.

It seems to have been introduced with #1216. That states an increased dependency on `setuptools` 61, but that seems not to be strictly necessary, nor make a difference.

Details as follows:

This is all running in a VM/container of Ubuntu 20.04 setup under Rancher Desktop on an M1 Mac Mini.

Initial software versions/environment:
```
# cat /etc/issue
Ubuntu 20.04.6 LTS

# uname -a
Linux bb8e17fb54db 6.1.30-0-virt #1-Alpine SMP Fri, 26 May 2023 06:53:59 +0000 aarch64 aarch64 aarch64 GNU/Linux

# python -m venv /srv/venvs/2023-07-05-netcdftest/

# source /srv/venvs/2023-07-05-netcdftest/bin/activate

# which python
/srv/venvs/2023-07-05-netcdftest/bin/python

# python --version
Python 3.8.10

# pip --version
pip 20.0.2 from /srv/venvs/2023-07-05-netcdftest/lib/python3.8/site-packages/pip (python 3.8)

# pip list
Package       Version
------------- -------
pip           20.0.2 
pkg-resources 0.0.0  
setuptools    44.0.0 

# dpkg -l lib{netcdf,hdf5}-dev
||/ Name           Version                 Architecture Description
ii  libhdf5-dev    1.10.4+repack-11ubuntu1 arm64        Hierarchical Data Format 5 (HDF5) - development files - serial version
ii  libnetcdf-dev  1:4.7.3-1               arm64        creation, access, and sharing of scientific data

# gcc --version
gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
```

Expected behavior:
```
# pip install netCDF4==1.6.2
Collecting netCDF4==1.6.2
  Using cached netCDF4-1.6.2.tar.gz (777 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
[...]

PR #1267, opened 2023-07-10
Add pyproject fmt

ocefpaf:
This should prevent a regression in https://github.com/Unidata/netcdf4-python/issues/1265 and complements @mwtoews  https://github.com/Unidata/netcdf4-python/pull/1266.

Note that the use of pre-commits has a learning curve and while I would advocate to a even heavier use than just this one, I'm OK if this PR is closed in case folks don't want to go there. However, if you do, one can set it up locally and run `pre-commit` or use the pre-commit hook. Additionally one can enable [pre-commit-ci](https://pre-commit.ci/) and use the `pre-commit.ci autofix` comment in the PR to fix it automatically. Enabling pre-commit-ci has the advantages of having this as a CI check and get the auto-updates for the hooks in the .pre-commit.yaml file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,110 +1,83 @@
 [build-system]
-requires = [
-    "Cython>=0.29",
-    "numpy>=2.0.0",
-    "setuptools>=77.0.1",
-    "setuptools_scm[toml]>=3.4",
-]
 build-backend = "setuptools.build_meta"
+requires = [
+  "cython>=0.29",
+  "numpy>=2",
+  "setuptools>=77.0.1",
+  "setuptools-scm[toml]>=3.4",
+]
 
 [project]
-name = "netCDF4"
+name = "netcdf4"
 description = "Provides an object-oriented python interface to the netCDF version 4 library"
-authors = [
-  {name = "Jeff Whitaker", email = "whitaker.jeffrey@gmail.com"},
-]
-requires-python = ">=3.10"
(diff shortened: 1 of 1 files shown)


ocefpaf:
Interesting, it broke what it was supposed to fix, the pyproject.toml! I'll try to figure out what is happening...

ocefpaf:
Found the problem and it may be a bug in the linter. xref.: https://github.com/tox-dev/pyproject-fmt/issues/110

ocefpaf:
There is an opendap sever failure now that is unrelated.

ocefpaf:
This one is ready for review.

mwtoews:
I've been increasing my use of pre-commit (+ .ci) lately, but I've not heard of pyproject-fmt until now. I'm unsure about it's inconsistency, however. For example, some sections have a 1-line space between and others don't, some have sorted keys, others don't. Some local tests that I tried with this branch, where I've re-arranged some items in the pyproject file will yield bad formats too, so I can't grasp what the tool is trying to do.

neutrinoceros:
`check-toml` would solve the issue of shipping invalid toml, without the risk of making it invalid itself: this hook only reads data, but doesn't write back

https://github.com/pre-commit/pre-commit-hooks/blob/e5e94e8702cbd993aaa4b67559f576dde7ef7adc/.pre-commit-hooks.yaml#L82

PR #1268, opened 2023-08-02
Try to improve error messages when creating `Variable`s

ZedThree:
Error messages are currently reported directly from the netcdf-C error string, which doesn't include information about the specific object. This can make debugging awkward or painful. This PR simply includes the variable and group names in error messages arising from `Variable.__init__`.

Minimal example:

```python
import netCDF4

with netCDF4.Dataset("badname.nc", "w", format="NETCDF4") as f:
    time = f.createDimension("time", None)
    g = f.createVariable("badname ", "i4", ("time",))
```

Error without this PR:

```
RuntimeError: NetCDF: Name contains illegal characters
```

With this PR:

```
RuntimeError: NetCDF: Name contains illegal characters: (variable 'badname ', group '/')
```

Exact format could be tweaked and used in more places.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,6 +1,7 @@
  version 1.6.5 (not yet released)
 ===============================
  * fix for issue #1271 (mask ignored if bool MA assinged to uint8 var)
+ * include information on specific object when reporting errors from netcdf-c
 
  version 1.6.4 (tag v1.6.4rel)
 ===============================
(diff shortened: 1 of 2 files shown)


ZedThree:
Note that unfortunately this likely can't be solved in the netCDF-C library itself because the error strings are `const char*`, so I don't think there's a good way to include dynamic information in the error message.

jswhit:
looks good - I can merge once you update to the current master and include a Changelog entry

PR #1277, opened 2023-09-29
Remove deprecated Cython `IF`, `DEF`

### LINKED ISSUE #1269: IF/DEF conditional compilation deprecated in cython 3.0
Cython will deprecate compile time conditionals, see https://github.com/cython/cython/issues/4310.

We use this quite extensively.

ZedThree:
Fixes #1269 

The main trick I've used here is having a new C header `netcdf-compat.h` which does two things:

1. moves most of the API checking out of `setup.py`
2. defines stubs for the missing functions when a feature is missing

This allows us to eliminate most of the `DEF` statements and compile-time `IF` conditionals. We can also include everything unconditionally from this compatibility header in `netCDF4.pxi`. The use of the functions is still guarded at runtime, and I've flipped most of these guards to raise exceptions as soon as possible. The stubs all return `NC_EINVAL` so if for some reason a guard is missed or bypassed, they'll still raise an error and not silently fail.

I needed to use a different technique for the conditional `cimport` of mpi4py. For this, I moved the two branches into separate files, `[no_]parallel_support_imports.pxi.in`. The build system then chooses which one to copy to `parallel_support_imports.pxi` which we can then unconditionally include in the `.pyx` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ dist/
 netCDF4/_netCDF4.c
 netCDF4/*.so
 include/constants.pyx
+include/parallel_support_imports.pxi
 netcdftime/_netcdftime.c
 venv/
 .eggs/
--- a/include/mpi-compat.h
+++ b/include/mpi-compat.h
@@ -4,6 +4,10 @@
 #ifndef MPI_COMPAT_H
 #define MPI_COMPAT_H
 
+#include "netcdf-compat.h"
+
+#if HAS_PARALLEL_SUPPORT
+
 #include <mpi.h>
 
 #if (MPI_VERSION < 3) && !defined(PyMPI_HAVE_MPI_Message)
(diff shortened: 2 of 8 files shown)


jswhit:
Thanks for working on this!  Since this is a pretty big change, and the IF/DEF support in cython is not going away anytime soon, I'd like to leave this open for discussion for a while before merging (after the next release perhaps).

ZedThree:
Sure, that sounds reasonable! I'm happy to break it into smaller PRs if you prefer

jswhit:
I don't see any reason to split it up, but we'll see what others think.

jswhit:
Now that 1.6.5 is out, this can be merged as soon as it is ready.  @ZedThree please bump the version number to 1.7.0 and add a Changelog entry.

ZedThree:
I think this is ready now

PR #1288, opened 2023-10-27
Simplify deeply nested conditionals for opening/creating file

ZedThree:
Basically, convert lots of the conditionals around opening the file into constructing the correct value of `cmode`. This leaves us with top-level conditionals for reading (with/without memory) and writing, and then one set of nested conditionals for serial/parallel.


jswhit:
thanks @ZedThree - I've been  meaning to clean this up for a long time

jswhit (approved):

Issue #1287, opened 2023-10-27
no python 3.7 wheels since netCDF4 1.6.5?


soxofaan:
Hi,

One of our CI runs (in github actions) started failing on py3.7 this week due to `pip install` failure
```
Collecting netCDF4
  Downloading netCDF4-1.6.5.tar.gz 
...
      ValueError: did not find HDF5 headers
```

I could track it down to netCDF4-1.6.5 not having built wheels for python 3.7 (https://pypi.org/project/netCDF4/1.6.5/#files)
Last release with wheels for py3.7 is https://pypi.org/project/netCDF4/1.6.4/#files

I understand that Python 3.7 end-of-lifed earlier this year, so I guess dropping python 3.7 support is intentional, however I can not find any mentioning of this in the changelog or news section of the readme.

soxofaan:
note that python 3.7 still looks supported in a couple of places:

https://github.com/Unidata/netcdf4-python/blob/df470a52c793df0da5b2939809162609ea0877d6/pyproject.toml#L15

In README.md:
> Make sure [numpy](http://www.numpy.org/) and [Cython](http://cython.org/) are installed and you have [Python](https://www.python.org/) 3.7 or newer.

jswhit:
we're not supporting python 3.7 anymore (numpy stopped supporting 3.7 in Dec 2021).  Those references in the docs need to be removed.

PR #1289, opened 2023-10-28
remove references to python 3.7 support

jswhit:
addresses issue #1287

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,6 +25,7 @@ classifiers = [
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -21,7 +21,6 @@ license = {text = "MIT"}
 classifiers = [
     "Development Status :: 3 - Alpha",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.7",
```
ocefpaf, line 24: Do you want to add 3.12 to the list? I have to confess I find classifiers not that useful and and super easy to become stale and outdated.

soxofaan:
I think dropping py3.7 support should also be mentioned in the changelog, e.g. under version 1.6.5, which is first version without py3.7 wheels on pypi

PR #1292, opened 2023-11-06
Allow running unit tests with pytest

ZedThree:
Some small changes to enable running the unit tests with either `unittest` or `pytest`

- use the `unittest.skipIf` decorator to skip tests, instead of removing them manually in `run_all.py`
- for the bundled test files, make sure the name is relative to the test directory
- rename the files to `test_*.py`

All of these mean running just `pytest` from the top level works, without breaking the existing testing workflow.


ZedThree:
This could go even further, of course, and rewrite everything the tests for `pytest`. Pytest doesn't need classes, it just looks for functions called `test_*`, and has some useful built-in fixtures.

jswhit (approved):

PR #1295, opened 2023-11-16
Add complex number support

### LINKED ISSUE #1280: Interest in extension for complex numbers?
I have a (currently proof-of-concept) [library extending netCDF to support complex numbers](https://github.com/PlasmaFAIR/nc-complex), including all of the most common conventions that people are currently using. This is implemented as a C library so that I can consistently extend all the language bindings. The Python API is built on top of netcdf4-python and is an entirely drop-in replacement/extension. Here's an example of writing a complex array and reading it back, demonstrating the correct dtype both ways:

```python
import nc_complex as netCDF4
import numpy as np

complex_array = np.array([0 + 0j, 1 + 0j, 0 + 1j, 1 + 1j, 0.25 + 0.75j], dtype="c16")

with netCDF4.Dataset(filename, "w") as f:
    f.createDimension("x", size=len(complex_array))
    complex_var = f.createVariable("complex_data", "c16", ("x",))
    complex_var[:] = complex_array

with netCDF4.Dataset(filename, "r") as f:
    print(f["complex_data"])
    print(f["complex_data"][:])

# <class 'nc_complex.Variable'>
# compound data_dim(x)
# compound data type: complex128
# unlimited dimensions:
# current shape = (5,)
# [0.  +0.j   1.  +0.j   0.  +1.j   1.  +1.j   0.25+0.75j]
```

Is there potentially any interest in merging the modifications for the Python API in here? It would require using my C library and switching out some calls to my wrappers. It could also be possible to make this behaviour optional if that was a concern.

As a side note, I found [scikit-build-core](https://scikit-build-core.readthedocs.io/en/latest/index.html) really, really useful for building the Cython module. Might simplify the build system here too?

ZedThree:
Closes #1280 

This adds support for reading/writing complex numbers in a variety of conventions, using my [nc-complex](https://github.com/PlasmaFAIR/nc-complex) library to do the heavy lifting. It's taken awhile because I had to go back and polish up the base library. It now supports writing complex numbers to netCDF3/classic model by automatically creating the complex dimension.

Complex number support is enabled by passing `auto_complex=True` to `Dataset`:

```python
import netCDF4
import numpy as np

complex_array = np.array([0 + 0j, 1 + 0j, 0 + 1j, 1 + 1j, 0.25 + 0.75j], dtype="c16")

filename = "complex.nc"
with netCDF4.Dataset(filename, "w", auto_complex=True) as f:
    f.createDimension("x", size=len(complex_array))
    c_var = f.createVariable("data", "c16", ("x",))
    c_var[:] = complex_array

with netCDF4.Dataset(filename, "r", auto_complex=True) as f:
    print(f["data"])

# <class 'netCDF4._netCDF4.Variable'>
# compound data(x)
# compound data type: complex128
# unlimited dimensions: 
# current shape = (5,)
```

Internally, `auto_complex` is then used to switch between calling the `nc_*` and `pfnc_*` versions of the various functions. `pfnc_def_var` is always used to create variables, as this just falls through to `nc_def_var` if it's not a complex datatype.

I still need to add an example and update the docs, but all the code is working and I've added some tests, so I thought I'd open the PR so you can see what it looks like and maybe try it out.

There's probably still some sharp edges around vlen or compound types with nested complex numbers. I've not tested them.


jswhit:
thanks @ZedThree!  I see you are adding a submodule that points to your repo - which is fine, but I think it should still be possible to build netcdf4-python without that submodule.  How hard would it be to add the ability to bypass the complex number support?

ZedThree:
> How hard would it be to add the ability to bypass the complex number support?

Definitely possible, but a little complicated. We'd need to do something like I added for the optional parallel support, switching out a `.pxi` header at build time. And then a runtime check if `auto_complex=True` but complex support is disabled.

The intention with the current implementation is that the complex number support is completely opt-in anyway. The underlying `nc-complex` C library is a single file and very cheap to build, so using the default `auto_complex=False` shouldn't add any overhead.

Happy to do this, but honestly I'm not sure it's worth making it optional at build-time.

> do you plan to try to get this integrated into netcdf-c?

From this discussion: https://github.com/Unidata/netcdf-c/discussions/2652 it seems unlikely to be accepted unfortunately.

ZedThree:
> in case your repo is broken or unavailable at build time

That was exactly my thinking for including it as a submodule, so that it would always be available :)

jswhit:
Looks good - guess all it needs is a Changelog entry

ZedThree:
Yep, will do! There's one more thing I want to add to the base library first, will get on that this week

ZedThree:
Ok, I think this is complete from my side

PR #1305, opened 2023-12-20
mpi-compat: support MPI_Message on MS-MPI

mathstuf:
MS-MPI does not define an accurate `MPI_VERSION`, so `MPI_Message` is mis-detected.

See: https://github.com/mpi4py/mpi4py/issues/19


jswhit:
Thanks! Please add a Changelog entry and I will merge.

PR #1307, opened 2024-01-09
Add custom build backend for parallel NetCDF4

stubbiali:
Introducing an [in-tree build backend](https://peps.python.org/pep-0517/#in-tree-build-backends) which programmatically detects whether NetCDF4 supports parallel I/O, and if so adds MPI4Py to the list of build dependencies. The custom backend is a very thin wrapper around the original Setuptools backend.


ZedThree:
You might want to report which `nc-config` it's trying to run. Currently the Windows builds are failing with `FileNotFoundError: [WinError 2] The system cannot find the file specified` which doesn't include the path, annoyingly!

stubbiali:
Many apologies, this PR has fallen out of my radar for too long... So far, CI on Windows fails since `nc-config` cannot be found. May be the PATH needs to be modified before trying to pip-install the package? Unfortunately, I don't have a Windows machine at disposal to do proper debug offline.

stubbiali:
~After the recent strip of commits (BTW thanks @jswhit for updating the branch!), our application codes using `netcdf4` now fail with `The MPI_Comm_dup() function was called before MPI_INIT was invoked`. The culprit seems to be [c07584f](https://github.com/Unidata/netcdf4-python/pull/1307/commits/c07584f72c427c9de599a18ea67be3bdc5ea314e). Any idea?~ Please ignore this comment, there were some issues in our setup, now everything works fine.

jswhit:
The windows wheel builds are failing because nc-config only works on linux and macos.

stubbiali:
Thanks for the clarification @jswhit and sorry for overlooking it. The code now relies on utilities fetched from `setup.py` to inspect if the C library has parallel support enabled. This should guarantee portability across all OS.

PR #1318, opened 2024-04-16
Fix bug in set_collective added in #1277

pgf:
Fix bug added in commit f64ee26773502c7568b1ceefe5967f83230f207c that cause `set_collective` to always set collective mode, regardless of the `value` input argument.


jswhit:
the mpi test with openmpi is failing with a "not enough slots" error.  Could you add --oversubscribe to the mpirun in .github/workflows/miniconda.yml (you just have to uncomment one line and comment out the line below it).

PR #1319, opened 2024-04-26
Test against np2

ocefpaf:
I'm only testing it on ubuntu-latest, x64, and py3.12, using conda packages it should be enough to catch if there are any problems with numpy 2.0.


ocefpaf:
I'm not sure why the mpi one is failing but the extra matrix item I added, experimental=true, is passing with numpy 2.0.

PR #1324, opened 2024-06-07
prepare for v1.7.0 release

jswhit:


headtr1ck:
Hey, do you think we can squeeze in #1302?
Would be fitting for a new minor release.

jswhit:
> Hey, do you think we can squeeze in #1302? Would be fitting for a new minor release.

Yes, this can go in 1.7.1

Issue #1329, opened 2024-06-13
netcdf4-1.7.0.tar.gz on PyPI is missing source files


cgohlke:
The latest source distribution on PyPI, [`netcdf4-1.7.0.tar.gz`](https://pypi.org/project/netCDF4/1.7.0/#files), is missing some source code files:

- `include/netcdf-compat.h`
- `include/parallel_support_imports.pxi.in`
- `include/no_parallel_support_imports.pxi.in`
- `external/nc_complex/*`

ocefpaf:
Indeed. Please see https://github.com/Unidata/netcdf4-python/issues/1331#issuecomment-2167315084.

aaraney:
I don't think the missing files in `include` are related to #1331. These just need to be added to the `MANIFEST.in` file.

```
include/netcdf-compat.h
include/parallel_support_imports.pxi.in
include/no_parallel_support_imports.pxi.in
```

ocefpaf:
Indeed, those 3 are not part of the `nc_complex`. I'll address those tomorrow morning.

aaraney:
@ocefpaf, no worries! I just opened https://github.com/Unidata/netcdf4-python/pull/1336 to add them!

mgorny:
Thanks. I can confirm that 1.7.1 looks good.

PR #1332, opened 2024-06-14
remove submodules from checkout actions

### LINKED ISSUE #1331: 1.7.0 fails to build (cc1: fatal error: external/nc_complex/src/nc_complex.c: No such file or directory)
The Debian package build for 1.7.0 failed to build due to missing generated source files:
```
I: pybuild plugin_pyproject:129: Building wheel for python3.12 with "build" module
I: pybuild base:311: python3.12 -m build --skip-dependency-check --no-isolation --wheel --outdir /build/netcdf4-python/.pybuild/cpython3_3.12  
* Building wheel...
-L/usr/lib/x86_64-linux-gnu -L/usr/lib/x86_64-linux-gnu/hdf5/serial -lnetcdf
reading from setup.cfg...
using nc-config...
checking /usr/include...
hdf5 headers not found in /usr/include
checking /usr/include/hdf5/serial...
HDF5 library version: 1.10.10 headers found in /usr/include/hdf5/serial
using netcdf library version b'4.9.2'
using Cython to compile netCDF4.pyx...
netcdf lib does not have parallel functions
NETCDF_PLUGIN_DIR not set, no netcdf compression plugins installed
running bdist_wheel
running build
running build_py
creating build
creating build/lib.linux-x86_64-cpython-312
creating build/lib.linux-x86_64-cpython-312/netCDF4
copying src/netCDF4/utils.py -> build/lib.linux-x86_64-cpython-312/netCDF4
copying src/netCDF4/__init__.py -> build/lib.linux-x86_64-cpython-312/netCDF4
running egg_info
creating src/netCDF4.egg-info
writing src/netCDF4.egg-info/PKG-INFO
writing dependency_links to src/netCDF4.egg-info/dependency_links.txt
writing entry points to src/netCDF4.egg-info/entry_points.txt
writing requirements to src/netCDF4.egg-info/requires.txt
writing top-level names to src/netCDF4.egg-info/top_level.txt
writing manifest file 'src/netCDF4.egg-info/SOURCES.txt'
reading manifest file 'src/netCDF4.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no previously-included files found matching 'examples/data'
warning: no previously-included files found matching 'src/netCDF4/_netCDF4.c'
adding license file 'LICENSE'
writing manifest file 'src/netCDF4.egg-info/SOURCES.txt'
copying src/netCDF4/_netCDF4.pyx -> build/lib.linux-x86_64-cpython-312/netCDF4
creating build/lib.linux-x86_64-cpython-312/netCDF4/plugins
copying src/netCDF4/plugins/empty.txt -> build/lib.linux-x86_64-cpython-312/netCDF4/plugins
running build_ext
Compiling src/netCDF4/_netCDF4.pyx because it changed.
[1/1] Cythonizing src/netCDF4/_netCDF4.pyx
building 'netCDF4._netCDF4' extension
creating build/temp.linux-x86_64-cpython-312
creating build/temp.linux-x86_64-cpython-312/external
creating build/temp.linux-x86_64-cpython-312/external/nc_complex
creating build/temp.linux-x86_64-cpython-312/external/nc_complex/src
creating build/temp.linux-x86_64-cpython-312/src
creating build/temp.linux-x86_64-cpython-312/src/netCDF4
x86_64-linux-gnu-gcc -fno-strict-overflow -Wsign-compare -DNDEBUG -g -O2 -Wall -g -O2 -Werror=implicit-function-declaration -ffile-prefix-map=/build/netcdf4-python=. -fstack-protector-strong -fstack-clash-protection -Wformat -Werror=format-security -fcf-protection -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION -DNC_COMPLEX_NO_EXPORT=1 -I/usr/include -I/usr/include/hdf5/serial -I/usr/lib/python3/dist-packages/numpy/core/include -Iinclude -Iexternal/nc_complex/include -Iexternal/nc_complex/include/generated_fallbacks -I/usr/include/python3.12 -c external/nc_complex/src/nc_complex.c -o build/temp.linux-x86_64-cpython-312/external/nc_complex/src/nc_complex.o
cc1: fatal error: external/nc_complex/src/nc_complex.c: No such file or directory
compilation terminated.
error: command '/usr/bin/x86_64-linux-gnu-gcc' failed with exit code 1

ERROR Backend subprocess exited when trying to invoke build_wheel
```
Full buildlog: [netcdf4-python.log](https://github.com/user-attachments/files/15831743/netcdf4-python.log)

ocefpaf:
@jswhit I'm not a big fan of submodules b/c they add an extra layer of complexity, specially with code tracking/versioning. This change will ship the latest submodule code with the sdist. We are already using the same when building the wheels. Which, BTW, stomped me for a while until I realized they were missing.

Closes #1331 and #1329.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -21,7 +21,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          submodules: 'true'
 
       - uses: actions/setup-python@v4
         name: Install Python
@@ -69,7 +68,6 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
-        submodules: 'true'
 
     # For aarch64 support
     # https://cibuildwheel.pypa.io/en/stable/faq/#emulation
@@ -130,7 +128,6 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          submodules: 'true'
 
       - uses: actions/setup-python@v4
         name: Install Python
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -21,6 +21,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          submodules: 'true'
 
       - uses: actions/setup-python@v4
         name: Install Python


sebastic:
sdist are not a good choice for packagers because maintaining MANIFEST.in is too error prone, that's why the Debian package uses the GitHub tar archives.

Vendoring the bits of nc-complex that are required to build is a better choice.

ocefpaf:
> sdist are not a good choice for packagers because maintaining MANIFEST.in is too error prone, that's why the Debian package uses the GitHub tar archives.

In a perfect GitHub world I agree with you but I've seeing so many bad practices on both PyPI sdist and GH tarballs that I don't think one is better than the other. I'm one of the core maintainers for conda-forge and we deal with tons of Python scientific packages that, due to some "magic" are only possible to build from PyPI or GH depending on the level of acrobatics the maintainers are doing. Regardless, the use of submodules here broke both and we need to fix it.

> Vendoring the bits of nc-complex that are required to build is a better choice.

Agreed. I'm not a main maintainer here. Let's see what they have to say. Maybe an alternative solution is to make that code optional. Not sure...

jswhit:
@ZedThree - having nc-complex as a submodule in this repo is causing some issues with releases (see also issue #1331).  Do you have any suggestions on how to fix this?

sebastic:
> Do you have any suggestions on how to fix this?

Vendor the three required files, e.g. apply this patch:

 https://salsa.debian.org/debian-gis-team/netcdf4-python/-/blob/debian/1.7.0-1/debian/patches/nc-complex.patch?ref_type=tags

dopplershift:
If you use setuptools_scm, you don't need to worry about `MANIFEST.in`.

jswhit:
> > Do you have any suggestions on how to fix this?
> 
> Vendor the three required files, e.g. apply this patch:
> 
> https://salsa.debian.org/debian-gis-team/netcdf4-python/-/blob/debian/1.7.0-1/debian/patches/nc-complex.patch?ref_type=tags

by 'vendoring' you mean just including a fork of the needed source files directly in the rep?

sebastic:
> by 'vendoring' you mean just including a fork of the needed source files directly in this repo?

Yes, just copy the files needed for netcdf4-python into its source tree. And copy them again when you would have updated the submodule for a newer version.

jswhit:
> If you use setuptools_scm, you don't need to worry about `MANIFEST.in`.

done in PR #1337

aaraney:
I would recommend continuing to use the submodule over a naive vendoring approach. You lose provenance, version, and licensing information (the last not so applicable to this project). While submodules can be a pain, updating vendored code is far more error prone and will be more difficult maintain in the long term. It will also be more difficult to determine where and when breaking changes occurred.

ocefpaf:
@aaraney while all that is true, the reasons why these where submodules in the first place are not strong enough to justify using it. Maybe this code should always have been part of the main code base.

I will modify this PR to remove the submodule init to conform with #1337.

aaraney:
> @aaraney while all that is true, the reasons why these where submodules in the first place are not strong enough to justify using it

Can you elaborate on why you think that is the case? I will also note, vendoring code in this way is also an increased security risk. It's much easier to verify where code came using submodules. A naive vendoring approach is much more susceptible to for example, a social engineering attack.

From other issues I've read, the main justification has been it's burdensome when building wheels in CI. This is remedied by initializing the submodules after the checkout step in the gh action and including the necessary files in `MANIFEST.in`.

sebastic:
> From other issues I've read, the main justification has been it's burdensome when building wheels in CI. This is remedied by initializing the submodules after the checkout step in the gh action and including the necessary files in `MANIFEST.in`.

It also makes it impossible for distributions to build packages using the GitHub tarball. `git archive` does not recursively include the submodules. PyPI sdists are not an acceptable alternative, it only provides the latest version whereas prior versions can be downloaded via the GitHub tags.

aaraney:
Thanks for the added context, @sebastic! That is really unfortunate... it looks like there have been several unsuccessful requests to add such a feature:

- https://github.com/orgs/community/discussions/31795
- https://github.com/orgs/community/discussions/20797
- https://github.com/AcademySoftwareFoundation/MaterialX/issues/317

I the spirit of trying to find a compromise it seems like there are two main options:
- Use a github action to add a release artifact that includes submodules. This is what `pytorch` does, for example ([PR introducing feature](https://github.com/pytorch/pytorch/pull/63022), [gh action workflow](https://github.com/pytorch/pytorch/blob/f0d68120f4e99ee6c05f1235d9b42a4524af39d5/.github/workflows/create_release.yml)).
- Vendor files from an external source (e.g. `git` repo) and use a tool in CI / local development to synchronize changes. For example, a tool like, [`vendir`](https://github.com/carvel-dev/vendir) or something hand rolled. The tool should be configured using checked in files that leave a paper trail to the source and a version of the vendored files.

I lean slightly to the second option b.c. fewer / no upstream changes should be required by for example, package managers. Not to mention it avoids the whole `xz` security conversation. My main concern is provenance and specifically version / tag tracking for transparency and debugging sake (e.g. debugging `netCDF4-python` builds in a, lets call it, unique environment 😄).

sebastic:
If nc-complex installs its headers and code as a header-only library it doesn't need to be vendored, nor when it provides the functions via a shared library.

aaraney:
> If nc-complex installs its headers and code as a header-only library it doesn't need to be vendored, nor when it provides the functions via a shared library.

If I am following correctly, you are talking about the case when `nc-complex` is already compiled and present on the system?

sebastic:
> > If nc-complex installs its headers and code as a header-only library it doesn't need to be vendored, nor when it provides the functions via a shared library.
> 
> If I am following correctly, you are talking about the case when `nc-complex` is already compiled and present on the system?

Yes.

I haven't dug into which bits of nc-complex are being used by netcdf4-python, and whether or not it's feasible to expect those to be exposed by public headers and functions, but in theory it would remove the need to vendor this code in netcdf4-python.

aaraney:
I see what you are saying. It could be treated like the `netcdf-c` and `netcdf-fortran` dependencies. I assume that's not being done because `nc-complex` is not readily available on package indexes?

ocefpaf:
This one will pas only after #1337 is merged.

jswhit:
> This one will pass only after #1337 is merged.

PR #1337 is merged.

PR #1336, opened 2024-06-14
chore: include missing netcdf headers in manifest / sdist

aaraney:
Include missing `netcdf` headers during packaging (i.e. in an `sdist`). These are opened in `setup.py` in certain build configurations if building from a source distribution.

related #1329


jswhit:
this should be fixed by using setuptools_scm (part of PR #1337)

aaraney:
Yeah, looks like we have duplicated some work here. Given how small the scope of this change is relative to #1337 I will keep this open and let the maintainers decide what direction to go.

PR #1340, opened 2024-06-18
fix download-artifact

ocefpaf:
Should fix the failure we are seeing in https://github.com/Unidata/netcdf4-python/actions/runs/9556646601/job/26344131394

Sorry I missed this one.


jswhit:
Thanks @ocefpaf, I was going to ask you about that one.

Issue #1347, opened 2024-07-05
Expose nc_rc_set


magnusuMET:
* python-netcdf 1.6.5-2
* netcdf 4.9.2
* Arch Linux, python 3.12.3
* Can't call `nc_rc_set` to set custom configuration

In newer versions of `netCDF` one is required to explicitly set a path to e.g. `.netrc`, see https://github.com/Unidata/netcdf-c/issues/1966. I can't seem to find a way to call this function from python as this function is not exposed.

jswhit:
PR #1347 exposes `nc_rc_set` (via `rc_set`) and `nc_rc_get` (via `rc_get`).

For some reason, `rc_get` triggers a segfault (crashing the interpreter) if you ask for a non-existent key.

jswhit:
@magnusuMET would you be able to test PR #1348 to see if it resolves your issue?

magnusuMET:
As far as I can tell this sets the key appropriately. I am seeing an issue where the key is not used by netcdf-c, but I think that is an unrelated issue

jswhit:
closed by PR #1348

PR #1348, opened 2024-07-06
expose nc_rc_set, nc_rc_get (via rc_set, rc_get module functions).

jswhit:
addresses PR #1347


src/netCDF4/__init__.pyi
```diff
@@ -624,6 +625,8 @@ def chartostring(
 ) -> npt.NDArray[np.str_] | npt.NDArray[np.bytes_]: ...
 
 def getlibversion() -> str: ...
+def nc_get(key: str) -> str: ...
+def nc_set(key: str, val: str): ...
```
headtr1ck, line 629:
```suggestion
def nc_set(key: str, val: str) -> None: ...
```

src/netCDF4/__init__.pyi
```diff
@@ -624,6 +625,8 @@ def chartostring(
 ) -> npt.NDArray[np.str_] | npt.NDArray[np.bytes_]: ...
 
 def getlibversion() -> str: ...
+def nc_get(key: str) -> str: ...
```
headtr1ck, line 628:
```suggestion
def rc_get(key: str) -> str: ...
```
If I am not mistaken. Same for set.

headtr1ck (commented):
You need to import the newly added things in _netCDF4.pyi to fix the stubtest

src/netCDF4/__init__.pyi
```diff
@@ -624,6 +625,8 @@ def chartostring(
 ) -> npt.NDArray[np.str_] | npt.NDArray[np.bytes_]: ...
 
 def getlibversion() -> str: ...
+def rc_get(key: str) -> str: ...
```
mulimoen, line 628: I think this signature should be
```suggestion
def rc_get(key: str) -> str | None: ...
```
and a null pointer check is necessary

headtr1ck: I don't see where it should return None.

jswhit: rc_get now checks for null pointer, returns None

PR #1349, opened 2024-07-12
Corrections and enchancements to type stubs

RandallPittmanOrSt:
This PR includes some corrections and enhancements to the type stubs provided in #1302. Some test cases and examples were modified during the process of testing the stubs on them with `mypy`.

## Enhancements

- Reorganized all the type literals into a meaningful hierarchy.
- `Variable.dtype` and `Variable.datatype` properties are overloaded via faux descriptor class workaround (ref. these [mypy](https://github.com/python/mypy/issues/9937#issuecomment-1599150232) and [pyright](https://github.com/microsoft/pyright/issues/3071#issuecomment-1043978070) issues). This is the only way we can actually get correct types out of these in static analysis.
- mypy checks in CI have been made more exhaustive by using `--check-untyped-defs`. (Some nuisance errors had to be silenced with the `--allow-redefinition` flag and some `# type: ignore` comments.)
- Allow more verbosity in `test/run_all.py`

## Fixes

- To stubs:
  - All function/method arguments that accept literals also need to accept the base types (e.g. `str` or `int`). It may be argued that this renders inclusion of the type literals (`'u2'`, `'i4'`, etc.) moot but they are left in as they can be useful in IDE pop-up documentation.
  - Move `dtype_is_complex` into `__init__.pyi`. It is publicly accessible in `netCDF4` due to import of `__all__` from `netCDF4._netCDF4`.
  - ~The `GetSetItemKey` for type alias `__getitem__` and `__setitem__` needs to also accept `float` (might be a mathematical integer) and `str` (might be convertible to integer) and also needs separate `list[]` generics defined for each possible type, rather than `list[int | bool | etc...]` due to lists be invariant and mutable.~ EDIT: We decided to accept anything for `key`.
  - `Dimension` can be created with `float` size if it's a mathematical integer.
  - `zlib`, `shuffle`, `fletcher32`, and `contiguous` arguments for creating a variable don't have to be `bool`, just "truthy", so they are changed to `Any`.
  - `Variable` typing and overloading
    - The `TypeVar` for `Variable` (now called `VarT`) has been changed to be unbound. The types in the old `bound=` argument weren't really correct nor useful.
    - There are some overloads for `Variable.__new__` (and `createVariable`) where if a variable is created using a NumPy type or `str` as the datatype, then the type of the variable can be known.
    - The type of a variable created with a datatype other than `str` or a NumPy type cannot be defined statically. In the future, we might be able to make `CompoundType`, `EnumType`, and/or `VLType` generic in which case we could possibly use those specified generics as types for `Variable`.
  - When creating an `EnumType, the `enum_dict` arg has been changed to `Mapping` instead of `dict`... This helps with [co/in]variance issues.
  - `Dataset.__dealloc__` is a Cython-only method and not publically accessible, so it was removed from the stubs.
- Various minor fixes made to test cases based on new mypy runs
  - `test_enum.EnumDictTestCase` wasn't called due to bad indentation of methods. Fixed some issues with that test case.
  - Use `ma.masked_array` instead of `ma.core.MaskedArray`. NumPy has gone back and forth with `MaskedArray` being part of `ma.core` or just `ma` depending on NumPy version. `ma.masked_array` appears to be a reliable alias in all cases.
  - Add some `isinstance` assertions for type narrowing
  - Add a couple of assertions where something wasn't actually being tested
    - `test_masked.py` - around line 96
    - `test_multifile.py` - around line 77
  - Rename some variables where the same name was being used with different meanings in adjacent contexts. (`test_multifile.py`)

## "Cosmetic" changes

- Reformatted with `ruff`. Used 130-character limit as per <https://typing.readthedocs.io/en/latest/source/stubs.html#maximum-line-length>. To see changes in this PR excluding this reformat, see here: <https://github.com/RandallPittmanOrSt/netcdf4-python/compare/851d453...rwp_typing>.
- Renamed a number of `TypeAlias`es to be singular where each represents a single thing, not a collection.

## Future possible changes

It might be better to allow any string or int where we're requiring `Literal` arguments in functions, while still providing the `Literal`s for auto-completion tooling and documentation. Many of the modifications to the test cases could be reverted if we did this.


jswhit:
@headtr1ck would you be able to review this?  If so, you should have a github invite to enroll as a collaborator.  After you accept, and I can add you as a reviewer.

src/netCDF4/__init__.pyi
```diff
@@ -473,15 +606,14 @@ class Variable(Generic[T_Datatype]):
...
     def __delattr__(self, name: str) -> None: ...
     def __setattr__(self, name: str, value: Any) -> None: ...
     def __getattr__(self, name: str) -> Any: ...
-    def __getitem__(self, elem: GetSetItemKey) -> np.ndarray: ...
+    def __getitem__(self, elem: GetSetItemKey) -> Any: ...
     def __setitem__(self, elem: GetSetItemKey, data: npt.ArrayLike) -> None: ...
     def __array__(self) -> np.ndarray: ...
     def __len__(self) -> int: ...
-
+    def __iter__(self) -> Iterator[Any]: ...  # faux method so mypy believes Variable is iterable
```
headtr1ck, line 616: Why is this required?

headtr1ck (commented):
Lots of unrelated changes in tests, but they seem reasonable.

.github/workflows/build_master.yml
```diff
@@ -83,5 +83,5 @@ jobs:
     - name: Stubtest
       run: |
         stubtest netCDF4 --allowlist .github/stubtest-allowlist --mypy-config-file=pyproject.toml
-        mypy test
-        mypy examples
+        mypy test --check-untyped-defs --allow-redefinition
```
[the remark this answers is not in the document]
RandallPittmanOrSt: Will do.

src/netCDF4/__init__.pyi
RandallPittmanOrSt: Because otherwise you get errors from mypy if you try to iterate over a Variable. [`mypy` doesn't understand that if an object as `__len__` and `__getitem__` then it is iterable](https://stackoverflow.com/questions/61712618/why-does-mypy-not-consider-a-class-as-iterable-if-it-has-len-and-getitem). This came up when I added the `--check-untyped-defs` flag to mypy to look more in-depth into the tests and examples.

RandallPittmanOrSt:
> Lots of unrelated changes in tests, but they seem reasonable.

They mostly came up when I started running mypy on the tests and examples with `--allow-untyped-defs` (or using pyright in VSCode to inspect the code).

src/netCDF4/__init__.pyi
```diff
@@ -197,80 +296,102 @@ class Dataset:
...
-        shuffle: bool = True,
-        szip_coding: Literal['nn', 'ec'] = 'nn',
-        szip_pixels_per_block: Literal[4, 8, 16, 32] = 8,
-        blosc_shuffle: Literal[0, 1, 2] = 1,
-        fletcher32: bool = False,
-        contiguous: bool = False,
+        datatype: np.dtype[NumericVarT] | type[NumericVarT],
+        dimensions: DimensionsSpecifier = (),
+        compression: CompressionType | str | None = None,
+        zlib: Any = False,
```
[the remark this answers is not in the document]
RandallPittmanOrSt: Good question. My reasoning was that in practice those parameters allow anything "truthy". I guess the overall principles I eventually came to were to try to avoid false positives as much as possible for people who start using the stubs, even if they're doing something a little weird like using an int or `None` for truthy parameter. This is at odds with type-stubs being a form of self-documentation. I'm open to reverting them back to `bool`.

src/netCDF4/__init__.pyi
```diff
@@ -1,23 +1,80 @@
+"""__init__.pyi - Type stubs for the netCDF4 Python package"""
+# Notes:
+#
+# - The stubs in this file are manually-generated and must be updated if and when the API is changed.
+# - The following **ruff** commands may be used to properly format this file according to
+#   https://typing.readthedocs.io/en/latest/source/stubs.html
+#
+#   ruff format --line-length 130 src/netCDF4/__init__.pyi  # format code
+#   ruff check --line-length 130 --select I --fix src/netCDF4/__init__.pyi  # sort imports
```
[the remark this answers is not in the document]
RandallPittmanOrSt: I don't think it's possible to only specify ruff configuration for certain files in `pyproject.toml`. While the typing docs recommend 130 char line length for stub files, I don't know if maintainers would want that length for `.py` files. Then again, this project currently doesn't have any auto-formatting, so maybe this is all moot. The comment at the top of this file was just a recommendation for how to work on the stubs file, independent of the rest of the project.

src/netCDF4/__init__.pyi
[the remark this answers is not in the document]
headtr1ck: Just because you can doesn't mean you should.
I think bool is the better type hint here, this is widely adopted by other projects.

test/test_utils.py
```diff
@@ -303,7 +303,7 @@ def test_ellipsis(self):
         assert_equal(take_ind[0,0,0,0,0], (slice(None), slice(None), slice(None), slice(None), slice(None)))
 
         try:
-            elem=(Ellipsis, [15,16,17,18,19], slice(None))
+            elem=(Ellipsis, [15,16,17,18,19], slice(None))  # type: ignore  # mypy doesn't like redefinition of elem here
```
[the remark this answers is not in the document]
RandallPittmanOrSt: Yeah, I'll just go ahead and do that where needed. Thanks.

src/netCDF4/__init__.pyi
```diff
@@ -1,23 +1,80 @@
...
+#   - Automatic typing of a Variable on variable creation is tedious due to the large number of ways to specify a variable's type
+#     (in particular all the type literals).
+#   - It is not possible to statically type a Variable of any user-defined type (CompoundType, EnumType, VLType) as these types
+#     are created dynamically.
+#   It is thus best left to the user to implement TypeGuards and/or perform other mixed static/runtime type-checking to ensure the
+#   type and shape of data retrieved from this library.
+# - `Dataset.__getitem__()` may return either a Variable or a Group, depending on the string passed to it. Rather than return a
+#   Union of Variable and Group, the authors of these stubs have elected to to return Any, leaving it up to users to determine the
+#   type of the returned value.
+# - `MFDataset.dimensions` returns `dict[str, Dimension]` and `MFDataset.variables` returns `dict[str, Variable]` even though the
```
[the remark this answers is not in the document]
RandallPittmanOrSt: Yeah, the implementation is... weird. If you think we should do anything different with this in the stubs, LMK. I'm not ready to jump in to refactoring the Cython code.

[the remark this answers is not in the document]
headtr1ck: This would be definitely out of scope. Lets leave it like it is.
Possibly, we could create a new issue for the `_Dimension` refactoring.

test/test_utils.py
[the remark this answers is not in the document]
RandallPittmanOrSt: In 620faf73

.github/workflows/build_master.yml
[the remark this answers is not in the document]
RandallPittmanOrSt: In 252deab1

src/netCDF4/__init__.pyi
headtr1ck:
```suggestion
    def __iter__(self) -> Iterator[np.ma.masked_array]: ...  # faux method so mypy believes Variable is iterable
```

src/netCDF4/__init__.pyi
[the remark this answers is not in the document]
RandallPittmanOrSt: Done in 31513eec.

src/netCDF4/__init__.pyi
```diff
@@ -541,14 +541,14 @@ class Variable(Generic[VarT]):
...
+        zlib: bool = False,
         complevel: CompressionLevel | int | None = 4,
-        shuffle: Any = True,
-        szip_coding: Literal["nn", "ec"] = "nn",
-        szip_pixels_per_block: Literal[4, 8, 16, 32] = 8,
-        blosc_shuffle: Literal[0, 1, 2] = 1,
-        fletcher32: Any = False,
-        contiguous: Any = False,
+        shuffle: bool = True,
+        szip_coding: Literal["nn", "ec"] | str = "nn",
```
headtr1ck, line 547: Do you think that it occurs often that people somehow pass this dynamically instead of directly in the `__init__`?
I know that it probably happens in tests, but in real-world scenarios I doubt that leaving it only as literal will cause problems.

src/netCDF4/__init__.pyi
RandallPittmanOrSt: Not necessarily. It depends on whether `always_mask` is set (could just return an `ndarray` ), or if it's a string variable then it may just return a string. Same thing as `__getitem__` and `getValue`.

src/netCDF4/__init__.pyi
RandallPittmanOrSt: It's precisely errors I got in the tests that caused me to make those changes. I don't know what the right answer is here. If you think it's better to only allow literals in the stubs I could be convinced.

headtr1ck: looking at other projects I see that the str union is only added if the literals are dynamically extended by further projects or are simply too many. If it is not too much work, I would prefer to keep the literal only (But no strong preference).

RandallPittmanOrSt:
I have changes coming for `GetSetItemIdx` based on testing I've been doing (see issue #1350). @headtr1ck  and @jswhit, I'd appreciate your thoughts on whether type stubs should allow \[sequences of\] floats and strings as variable indexes, irrespective of what the implementation actually allows.

headtr1ck:
> I have changes coming for `GetSetItemKey` based on testing I've been doing (see issue #1350). @headtr1ck and @jswhit, I'd appreciate your thoughts on whether type stubs should allow [sequences of] floats and strings as variable indexes, irrespective of what the implementation actually allows.

I noticed, that you can also use e.g. np.ndarrays for indexing. e.g. `var[[1, 2]]` is the same as `var[np.array([1, 2])]`. But not all iterables are supported, e.g. `var[{1, 2}]` fails.

RandallPittmanOrSt:
I needed to create a bunch of `TypeGuard` for tests so I put them in a helper module `test/type_guard.py`. mypy got confused with about importing `type_guard` so I had to add some funny lines to the mypy config in `pyproject.toml`. Probably a better approach would be revamping the tests a bit...

I thought about making `type_guard.py` a module in the actual distributed source and just calling it `typing.py`... then you could import it with something like

```python
import netCDF4.typing as n4t
```

a la `numpy.typing`. But I decided against it (for now, anyway).

RandallPittmanOrSt:
> I have changes coming for `GetSetItemKey` based on testing I've been doing (see issue #1350). @headtr1ck and @jswhit, I'd appreciate your thoughts on whether type stubs should allow [sequences of] floats and strings as variable indexes, irrespective of what the implementation actually allows.

These changes are now in. I decided not to allow a singular `bool` as a key (even though it's allowed at runtime), as this is a logical error, but I did allow all strings and floats. I decided not to allow mixed lists/Sequences of ints and floats, but if someone's linter complains down the road I guess that could be changed. But the TypeAlias was already long enough... perhaps the only other sane thing would be just go back to `Any`.

```python
GetSetItemKey: TypeAlias = (
    int
    | float
    | np.number
    | str
    | slice
    | ellipsis
    | list[int]
    | list[float]
    | list[np.number]
    | list[str]
    | list[bool]
    | list[np.bool_]
    | npt.NDArray[np.number]
    | npt.NDArray[np.bool_]
    | tuple[
        int
        | float
        | np.number
        | str
        | slice
        | ellipsis
        | Sequence[int]
        | Sequence[float]
        | Sequence[np.number]
        | Sequence[str]
        | Sequence[bool]
        | Sequence[np.bool_]
        | npt.NDArray[np.number]
        | npt.NDArray[np.bool_],
        ...,
    ]
)
```

RandallPittmanOrSt:
I haven't had time to try to set up the whole CI environment on my own machine but I don't _think_ the failures are caused by this PR.

jswhit:
yes, the mpi test failure is due an update in the conda mpi package - nothing to do with this PR. Just awaiting @headtr1ck's review now.

src/netCDF4/__init__.pyi
```diff
@@ -280,152 +409,168 @@ class Dataset:
...
         shuffle: bool = True,
-        szip_coding: Literal['nn', 'ec'] = 'nn',
+        szip_coding: Literal["nn", "ec"] = "nn",
         szip_pixels_per_block: Literal[4, 8, 16, 32] = 8,
         blosc_shuffle: Literal[0, 1, 2] = 1,
         fletcher32: bool = False,
         contiguous: bool = False,
-        chunksizes: Sequence[int] | None = None,
-        endian: EndianOptions = 'native',
+        chunksizes: int | None = None,
```
headtr1ck, line 486: Are you sure that the sequence is not correct?

src/netCDF4/__init__.pyi
```diff
@@ -474,15 +615,14 @@ class Variable(Generic[T_Datatype]):
...
     def __delattr__(self, name: str) -> None: ...
     def __setattr__(self, name: str, value: Any) -> None: ...
     def __getattr__(self, name: str) -> Any: ...
-    def __getitem__(self, elem: GetSetItemKey) -> np.ndarray: ...
+    def __getitem__(self, elem: GetSetItemKey) -> Any: ...
     def __setitem__(self, elem: GetSetItemKey, data: npt.ArrayLike) -> None: ...
     def __array__(self) -> np.ndarray: ...
     def __len__(self) -> int: ...
-
+    def __iter__(self) -> Iterator[Any]: ...  # faux method so mypy believes Variable is iterable
```
headtr1ck, line 625: Can we specify the type a bit better here?

src/netCDF4/__init__.pyi
```diff
@@ -474,15 +615,14 @@ class Variable(Generic[T_Datatype]):
     def set_ncstring_attrs(self, ncstring_attrs: bool) -> None: ...
     def set_collective(self, value: bool) -> None: ...
     def get_dims(self) -> tuple[Dimension, ...]: ...
-
     def __delattr__(self, name: str) -> None: ...
     def __setattr__(self, name: str, value: Any) -> None: ...
     def __getattr__(self, name: str) -> Any: ...
-    def __getitem__(self, elem: GetSetItemKey) -> np.ndarray: ...
+    def __getitem__(self, elem: GetSetItemKey) -> Any: ...
```
headtr1ck, line 621: Why did you change the return type here?
I feel like having this as np.ndarray is quite advantageous for end user code.

headtr1ck (commented):
Some good improvements, but I got some questions :)

src/netCDF4/__init__.pyi
RandallPittmanOrSt: Primarily because of the case of `str`. If you index a singular string value, the return type is `str`. E.g.

```python
# given "strvar" of type str and shape (20, 3)
strvar[0, 0] = "abc"
value = strvar[0, 0]
assert type(value) == str  # OK
```

So the return type either has to be a union of the two (I don't think it's solvable with overloads) or `Any`.

RandallPittmanOrSt: Furthermore if `var.auto_mask` is `False`, this may return values and not arrays. E.g.

```python
# given variable intvar of type int64 and shape (20, 3)
intvar.set_auto_mask(False)
intvar[...] = np.arange(60).reshape(20, 3)
assert type(intvar[0, 0]) == np.int64  # OK
```

RandallPittmanOrSt: The dimensions in these examples don't really matter, they're just what I used when I was playing around with the idea.

src/netCDF4/__init__.pyi
RandallPittmanOrSt: Not easily, partly for the same reasoning as `__getitem__`, below. If the variable is 1-D, the iterator may return either a scalar or an array, depending on `auto_mask` (unless the variable is `str` type in which case it will always return a `str`). If it's multi-dimensional, it will return an array.

src/netCDF4/__init__.pyi
RandallPittmanOrSt: Good catch. This was an erroneous change. I've reverted it.

src/netCDF4/__init__.pyi
```diff
@@ -280,152 +409,168 @@ class Dataset:
...
+    @overload
+    def __get__(self, instance: Variable[RealVarT], owner: Any) -> RealVarT: ...
+    @overload
+    def __get__(self, instance: Variable[ComplexVarT], owner: Any) -> CompoundType: ...
+    @overload
+    def __get__(self, instance: Variable[str], owner: Any) -> VLType: ...
+    @overload
+    def __get__(
+        self, instance: Variable[Any], owner: Any
+    ) -> Any: ...  # actual return type np.dtype | CompoundType | VLType | EnumType
```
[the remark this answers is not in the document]
RandallPittmanOrSt: Good catch. For real-typed variables, `.datatype` returns a `dtype`. I've fixed this.

src/netCDF4/__init__.pyi
```diff
@@ -26,48 +86,119 @@ if sys.version_info >= (3, 10):
...
+    | type[int | float | NumPyNumericType | str | np.str_]
+    | NetCDFUDTClass
+)
+
+VarT = TypeVar("VarT")
+RealVarT = TypeVar("RealVarT", bound=NumPyRealType)
+ComplexVarT = TypeVar("ComplexVarT", bound=NumPyComplexType)
+NumericVarT = TypeVar("NumericVarT", bound=NumPyNumericType)
+
+DimensionsSpecifier: TypeAlias = Union[str, bytes, Dimension, Iterable[Union[str, bytes, Dimension]]]
```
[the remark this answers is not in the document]
RandallPittmanOrSt: Good point. That's weird. Fixed.

src/netCDF4/__init__.pyi
```diff
@@ -26,48 +86,119 @@ if sys.version_info >= (3, 10):
...
 ]
 BoolInt: TypeAlias = Literal[0, 1]
 
 DateTimeArray: TypeAlias = npt.NDArray[np.object_]
 """numpy array of datetime.datetime or cftime.datetime"""
 
+# - Allow singular float or str keys, but not lists (or sequences) of them
+# - `list` and `ndarray` are specifically listed for the 1-D case as Sequence is too general -- it includes
+#   tuples which are rather for multi-dimensional indexing.
 GetSetItemKey: TypeAlias = (
```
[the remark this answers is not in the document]
RandallPittmanOrSt: Yeah, you're not wrong that it's complicated. I am pretty confident that this is correct, though there are some edge cases that the library might allow that the stubs are disallowing (e.g. a list of mixed float/int, also oddballs I tried to document in #1350). That said, I'm willing to accept that that this was a learning exercise and just leave it as `Any`...

@jswhit  do you have a position?

src/netCDF4/__init__.pyi
```diff
@@ -26,48 +86,119 @@ if sys.version_info >= (3, 10):
...
-    'zlib', 'szip', 'zstd', 'blosc_lz','blosc_lz4', 
-    'blosc_lz4hc', 'blosc_zlib', 'blosc_zstd'
-]
-CompressionLevelOptions: TypeAlias = Literal[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
-AccessModeOptions: TypeAlias = Literal['r', 'w', 'r+', 'a', 'x', 'rs', 'ws', 'r+s', 'as']
-FormatOptions: TypeAlias = Literal[
-    'NETCDF4', 'NETCDF4_CLASSIC', 'NETCDF3_CLASSIC',
-    'NETCDF3_64BIT_OFFSET', 'NETCDF3_64BIT_DATA'
+# fmt: on
+ComplexTypeLiteral: TypeAlias = Literal["c8", "c16", "complex64", "complex128"]
```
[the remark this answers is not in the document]
RandallPittmanOrSt: I understand what you're saying. However there needs to be a distinction in the name that this is a Literal, i.e. a `str` with a fully-defined set of possible values, rather than a `type`. I'm open to other suggestions but I think `ComplexType` implies that a value of this type is a `type`, e.g. `np.complex32`.

test/test_compression_bzip2.py
```diff
@@ -3,13 +3,15 @@
...
 from filter_availability import no_plugins, has_bzip2_filter
+import type_guards
 
 ndim = 100000
 filename1 = tempfile.NamedTemporaryFile(suffix='.nc', delete=False).name
 filename2 = tempfile.NamedTemporaryFile(suffix='.nc', delete=False).name
 array = uniform(size=(ndim,))
 
 def write_netcdf(filename,dtype='f8',complevel=6):
+    assert type_guards.valid_complevel(complevel) or complevel is None
```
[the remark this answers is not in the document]
RandallPittmanOrSt: You're right. I've updated things to be simpler.

examples/mpi_example.py
```diff
@@ -1,27 +1,34 @@
...
-    format = sys.argv[1]  # type: ignore
+from typing_extensions import TypeGuard
+if TYPE_CHECKING:
+    from netCDF4 import Format as NCFormat
 else:
-	format = 'NETCDF4_CLASSIC'
+    NCFormat = object
+
+
+def is_nc_format(fmt: str) -> TypeGuard[NCFormat]:
```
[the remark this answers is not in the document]
RandallPittmanOrSt: I agree. I've removed this stuff.

test/test_compression_blosc.py
```diff
@@ -1,8 +1,10 @@
 from numpy.random.mtrand import uniform
 from netCDF4 import Dataset
+import type_guards
 from numpy.testing import assert_almost_equal
 import os, tempfile, unittest, sys
 from filter_availability import no_plugins, has_blosc_filter
+import type_guards
```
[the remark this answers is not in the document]
RandallPittmanOrSt: Yup. Fixed with removal of `type_guards`.

src/netCDF4/__init__.pyi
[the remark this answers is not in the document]
jswhit: let's just go with Any and let the library raise an error if it gets something it doesn't like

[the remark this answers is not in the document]
RandallPittmanOrSt: Done.

jswhit:
@headtr1ck are you ok with this PR now?

jswhit:
Would like to make a 1.7.2 release this week, and include this if possible.  @headtr1ck could you please complete your review?

examples/parallel_test.nc
headtr1ck, line 1: Is there a reason this file has changed?
If not, maybe revert this change to be safe.
Same for `parallel_test_compressed.nc`

test/test_compression.py
```diff
@@ -30,8 +35,8 @@ def write_netcdf(filename,zlib,least_significant_digit,data,dtype='f8',shuffle=F
         #compression=''
         #compression=0
         #compression='gzip' # should fail
-    foo2 = file.createVariable('data2',\
-            dtype,('n'),compression=compression,least_significant_digit=least_significant_digit,\
+    foo2 = file.createVariable('data2',
+            dtype,('n'),compression=compression,least_significant_digit=least_significant_digit,  # type: ignore  # mypy doesn't like compression
```
headtr1ck, line 39: could add this to line 29:
```python
compression: CompressionLevel
if zlib:
    compression='zlib'
else ...
```

headtr1ck: thinking about it... maybe we should add an `int` Union to the `CompressionLevel` definition?
Keep the Literals for documentation.

On the other hand... how many people will actually not use this hardcoded but in a variable?

test/test_enum.py
```diff
@@ -64,26 +66,30 @@ def runTest(self):
...
             ev[...] = self.STORED_VAL
-        def tearDown(self):
-            os.remove(self.file)
-        def runTest(self):
-            with netCDF4.Dataset(file, 'r') as nc:
-                read_var = nc['evar']
-                assert read_var[...] == self.STORED_VAL 
-                assert read_et.enum_dict == self.VAL_MAP 
+
+    def tearDown(self):
```
headtr1ck, line 84: how was this even working before? haha.

headtr1ck (approved):
Besides the changed test files this looks good!
Sorry for the delay in review :)

headtr1ck:
I guess we have to solve some merge conflicts.
The test failures seems quite random and not related.

examples/parallel_test.nc
RandallPittmanOrSt: My bad. They're not in upstream and I accidentally committed them. I've removed them.

test/test_enum.py
RandallPittmanOrSt: I don't think it was being run due to bad indentation.

test/test_compression.py
RandallPittmanOrSt: > thinking about it... maybe we should add an `int` Union to the `CompressionLevel` definition? Keep the Literals for documentation.
> 
> On the other hand... how many people will actually not use this hardcoded but in a variable?

You bring up a good point.

I think this applies broadly to all the `Literal`s. On the side of making the Union, you get less noise but can miss bad literals (though including the `Literal` can still help with code completion). On the side of only allowing a literal is the most assurance that the type is correct, but with more potential noise/onus on the user to do runtime coercion to literals. 

And unfortunately, the type checkers don't support using something like 

```python
assert var in ["abc", "def"]
```

to coerce strings to literals.

I'm willing to make this change now, or wait until the next version if there are enough complaints about the noise in their type checkers.

headtr1ck: Yeah, my guess is that >95% of people will use inlined expressions for these kwargs or not even use them at all.

So let's wait until someone complains.

headtr1ck:
@jswhit looks good now!
Thanks!

jswhit:
@RandallPittmanOrSt if you update your fork to the latest develop I can merge

RandallPittmanOrSt:
@jswhit Done.

jswhit:
OK thanks @RandallPittmanOrSt - merging now

PR #1353, opened 2024-07-25
[docs] Add link to auth page on rc files.

RandallPittmanOrSt:
Incidentally pdoc re-inserted missing date2index and date2num stuff.

This is in the spirit of people seeing the functions and having no idea what they are for. I had to do a lot of digging since I was unfamiliar with this corner of the library, and in turn found out that the proper docs page regarding .rc files was missing (sort of) from the netcdf-c documentation (see <https://github.com/Unidata/netcdf-c/issues/2952>).

Could possibly wait until netcdf-c 4.9.3 is released, due to the currently-online netcdf-c docs page being out-of-date, but it should be updated soon.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1173,7 +1173,8 @@ returned.</p>
 <span>def <span class="ident">date2index</span></span>(<span>dates, nctime, calendar=None, select='exact', has_year_zero=None)</span>
 </code></dt>
 <dd>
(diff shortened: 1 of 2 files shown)


RandallPittmanOrSt:
Hold -- I'm double-checking with netcdf-c that I put in the correct URL (in the past it was sometimes "md_auth.html")

jswhit:
the index.html is auto-generated by pdoc from the docstrings in _netCDF4.pyx

PR #1359, opened 2024-08-13
Remove unused readme

ocefpaf:
Minor typos and removed an outdated readme file.


ocefpaf:
I'm not sure that the mpi error is but the Linux wheels are failing  b/c of Python 3.13 RC. We can either skip those, or leave as is and only modify if/when we need to build the other wheels. Leaving may help us diagnose things are other packages are built against it, like numpy.

PR #1377, opened 2024-10-23
Add py313 to the Windows wheel builds

ocefpaf:
@jswhit sorry but I forget to update this part here. I would do a manual upload to avoid a post release, we don't need to re-upload everything just for a single missing wheel.


ocefpaf:
The wheel from this run is in https://github.com/Unidata/netcdf4-python/actions/runs/11474274066. Here is a direct link https://github.com/Unidata/netcdf4-python/actions/runs/11474274066/artifacts/2091912527. Someone with access to PyPI will have to download, extract, and upload the py313 wheel.

jswhit:
windows py313 wheel manually uploaded

PR #1380, opened 2024-10-29
Set HTTP.NETRC if .netrc file found

### LINKED ISSUE #1299: netcdf > 1.6.1 does not read .netrc / .dodsrc
Hi,

To access CMEMS data-sources we use .netrc and .dodsrc, however it seems that versions post netcdf-1.6.1 no longer reads these files. Downgrading works fine. We use python-netcdf4 from conda-forge.

Best regards, Gaute

Relevant issue: https://github.com/OpenDrift/opendrift/issues/1083

magnusuMET:
Fixes #1299

### DIFF
--- a/src/netCDF4/_netCDF4.pyx
+++ b/src/netCDF4/_netCDF4.pyx
@@ -1322,7 +1322,7 @@ if __has_nc_rc_set__:
 if __has_nc_rc_set__:
     for path in [pathlib.Path(".netrc"), pathlib.Path.home().joinpath(".netrc")]:
         if path.exists():
-            if nc_rc_set("HTTP.NETRC", _strencode(path)) != 0:
+            if nc_rc_set("HTTP.NETRC", bytes(path)) != 0:
                 raise RuntimeError('error setting path to .netrc file')
             break
 
--- a/src/netCDF4/_netCDF4.pyx
+++ b/src/netCDF4/_netCDF4.pyx
@@ -1318,6 +1318,14 @@ if __has_nc_rc_set__:
     if nc_rc_set("HTTP.SSL.CAINFO", _strencode(certifi.where())) != 0:
         raise RuntimeError('error setting path to SSL certificates')
 
+# set .netrc if availble in cwd or home (issue #1299)
+if __has_nc_rc_set__:
+    for path in [pathlib.Path(".netrc"), pathlib.Path.home().joinpath(".netrc")]:
+        if path.exists():
+            if nc_rc_set("HTTP.NETRC", _strencode(path)) != 0:
+                raise RuntimeError('error setting path to .netrc file')
+            break
+
 def rc_get(key):
     """
 **```rc_get(key)```**


magnusuMET:
I think this is a new curl default as `netcdf-c` does not set any option related to the `.netrc` file. A better workaround might be to use `.ncrc` which can point to `.netrc`

jswhit:
@DennisHeimbigner would appreciate your input on this

DennisHeimbigner:
As noted, since using .netrc is off by default, it is necessary to have the HTTP.NETRC=&lt;netrc file path&gt;
set in .ncrc. Is this not working, or do you have a suggestion about changing this?

magnusuMET:
In versions prior to 1.6.1 `.netrc` was read by default and it broke on newer versions. I can't see any documentation for `.netrc` being read by default so this was an undocumented feature. I'd be fine with setting `HTTP.NETRC` in `.ncrc`

jswhit:
@DennisHeimbigner maybe netcdf-c needs to set CURLOPT_NETRC to  CURL_NETRC_OPTIONAL (looks like it is now set to CURL_NETRC_IGNORED by default on most systems)?   Maybe if it is set to  CURL_NETRC_IGNORED then setting HTTP.NETRC in .ncrc won't do anything.

jswhit:
@magnusuMET can you provide a reproducible example of the problem that this PR fixes? (including server URL and auth info)

DennisHeimbigner:
You are correct that the default is CURLOPT_NETRC_IGNORED.
If you specify a .netrc file using HTTP.NETRC, then CURLOPT_NETRC_OPTIONAL is  used.
We could certainly change the default to be CURLOPT_NETRC_OPTIONAL, although
I am not sure what the consequences would be.

magnusuMET:
@jswhit 
Server:
```shell
python3 -m http.server
```
Client code:
```python
import netCDF4
url = "http://localhost:8000/"
netCDF4.Dataset(url)
``` 
Contents of `.netrc`:
```
machine localhost
    login foo
    password bar
```
Run the client code with envionment `CURLOPT_VERBOSE=1`. When `.netrc` is present it should have `Authorization: Basic Zm9vOmJhcg==` in the output. The current version `python-netcdf` does not contain cause this line to be emitted, and auth is not set. In any case the program will error since the server does not speak DAP

jswhit2:
I don't see that "Authorization" message, instead  I get
```
* Host localhost:8000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
*   Trying [::1]:8000...
* Connected to localhost (::1) port 8000
* using HTTP/1.x
> GET /.dds HTTP/1.1
Host: localhost:8000
User-Agent: oc4.9.2
Accept: */*
Accept-Encoding: deflate, gzip, zstd

* Request completely sent off
* HTTP 1.0, assume close after body
< HTTP/1.0 404 File not found
< Server: SimpleHTTP/0.6 Python/3.12.6
< Date: Fri, 01 Nov 2024 22:07:36 GMT
< Connection: close
< Content-Type: text/html;charset=utf-8
< Content-Length: 335
<
* shutting down connection #0
syntax error, unexpected WORD_WORD, expecting SCAN_ATTR or SCAN_DATASET or SCAN_ERROR
context: <!DOCTYPE^ HTML><html lang="en"> <head> <meta charset="utf-8"> <title>Error response</title> </head> <body> <h1>Error response</h1> <p>Error code: 404</p> <p>Message: File not found.</p> <p>Error code explanation: 404 - Nothing matches the given URI.</p> </body></html>
  File "/Users/jwhitaker/test_netrc.py", line 3, in <module>
    netCDF4.Dataset(url)
  File "src/netCDF4/_netCDF4.pyx", line 2529, in netCDF4._netCDF4.Dataset.__init__
  File "src/netCDF4/_netCDF4.pyx", line 2166, in netCDF4._netCDF4._ensure_nc_success
OSError: [Errno -90] NetCDF: file not found: 'http://localhost:8000/
```
with or without the .netrc file present (using the mods in this PR).

jswhit2:
BTW, running _strencode on the Path object does not return bytes, I had to change that to from `_strencode(path)` to `bytes(path)`

magnusuMET:
Since this was an undocumented feature of `netcdf-c` I agree the inclusion of this might be suprising for users. If this change is wanted I can instead open a PR against `netcdf-c`

DennisHeimbigner:
It is documented in docs/auth.md

magnusuMET:
@DennisHeimbigner I see the following line: 
> If not specified, then libcurl will look first in the current directory, and then in the HOME directory.

I suppose this is an issue with `netcdf-c` and will raise an issue there

DennisHeimbigner:
Sorry I have lost track. Can you re-state the Issue to be fixed?

PR #1384, opened 2024-11-08
Disable Dataset iteration and membership operations (raise error on __iter__ and __contains__)

### LINKED ISSUE #1383: Attempts to iterate or perform a membership oepration on a Dataset returns a confusing error.
netCDF4 version: 1.7.2
Problem type: Usability

## Background Facts

1. `netCDF4.Dataset` is not meant to support iteration (`for item in dataset`) nor membership operations (`if item in dataset`). (Rather, one can do `for var in dataset.variables.items()` or `if varname in variables`, etc.)
2. If a class defines `__getitem__`, and does not define `__iter__` Python will basically turn the following code (where `container` is an instance of some class that defines `__getitem__`):

   ```python
   for item in container:
      ...
   ```

   into this:

   ```python
   i = 0
   while True:
       item = container[i]
       i += 1
       ...
   ```

3. If such a class does not contain a `__contains__` method, Python will basically turn this:

   ```python
   if val in container:
      ...
   ```

   into this:

   ```python
   if val in [item for item in container]:
      ...
   ```

   See <https://docs.python.org/3/reference/expressions.html#membership-test-operations>

## The problem

If a user tries to iterate on a dataset or perform a membership operation, Python attempts to send an integer argument to `Dataset.__getitem__`, and a confusing error results when in the path normalization step, something like this:

```text
File src/netCDF4/_netCDF4.pyx:2556, in netCDF4._netCDF4.Dataset.__getitem__()

File /shared/.pyenv/versions/3.9.16/lib/python3.9/posixpath.py:336, in normpath(path)
    334 def normpath(path):
    335     """Normalize path, eliminating double slashes, etc."""
--> 336     path = os.fspath(path)
    337     if isinstance(path, bytes):
    338         sep = b'/'

TypeError: expected str, bytes or os.PathLike object, not int
```

## Suggested Change

Explicitly define `__iter__` and `__contains__` and raise a `TypeError` when either is called.

RandallPittmanOrSt:
Fixes #1383.

With this change,

```py
for item in dataset:
    ...
```

raises `TypeError: Dataset is not iterable.`

Any

```py
if item in dataset:
    ...
```

raises `TypeError: Dataset does not support membership operations. Perhaps try 'varname in dataset.variables' or 'dimname in dataset.dimensions'.`


jswhit2:
Would be nice to support the iteration, but agree this is useful in the interim since the error message is much clearer.  Merging now...

PR #1386, opened 2024-11-13
DOC: More fixes to avoid &lt;code&gt; showing up in docs

DWesl:
I went and checked the docs produced by pdoc from #1382 and found three more instances of &lt;code&gt;, which I changed in the same way, and verified that `pdoc` produces the expected documentation.  Since the GitHub Pages workflow doesn't seem to re-generate `doc/index.html`, I decided to make the changes there as well.


DWesl:
Verified changes in docs now.  Thank you!

PR #1407, opened 2025-04-28
fix latest pip egg deprecation

ocefpaf:
Should fix #1406

### DIFF
--- a/.github/workflows/build_latest.yml
+++ b/.github/workflows/build_latest.yml
@@ -63,13 +63,14 @@ jobs:
     - name: Install python dependencies via pip
       run: |
         python -m pip install --upgrade pip
-        pip install numpy cython cftime pytest twine wheel check-manifest mpi4py typing-extensions
+        python -m pip install numpy cython cftime pytest twine wheel check-manifest mpi4py typing-extensions
 
     - name: Install netcdf4-python
       run: |
         export PATH=${NETCDF_DIR}/bin:${PATH} 
         export NETCDF_PLUGIN_DIR=${{ github.workspace }}/netcdf-c-${NETCDF_VERSION}/plugins/plugindir
-        python setup.py install
+        python -m pip install . --no-build-isolation
+
     - name: Test
       run: |
         export PATH=${NETCDF_DIR}/bin:${PATH} 
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,8 @@ from setuptools import setup, Extension
 from setuptools.dist import Distribution
 from typing import List
 
+
+
 open_kwargs = {'encoding': 'utf-8'}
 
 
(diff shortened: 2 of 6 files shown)


.github/workflows/build_latest.yml
```diff
@@ -63,13 +63,14 @@ jobs:
...
         python -m pip install --upgrade pip
-        pip install numpy cython cftime pytest twine wheel check-manifest mpi4py typing-extensions
+        python -m pip pip install numpy cython cftime pytest twine wheel check-manifest mpi4py typing-extensions
 
     - name: Install netcdf4-python
       run: |
         export PATH=${NETCDF_DIR}/bin:${PATH} 
         export NETCDF_PLUGIN_DIR=${{ github.workspace }}/netcdf-c-${NETCDF_VERSION}/plugins/plugindir
-        python setup.py install
+        python -m pip install .
```
ocefpaf, line 72: @jswhit this is the only part that we need to fix, but I confess I got lost with the parallel errors here. Any advice?

jswhit:
@ocefpaf do you mean this [one](https://github.com/Unidata/netcdf4-python/actions/runs/14711548396/job/41284945149)?

.github/workflows/build_latest.yml
jswhit: I think this is due to the fact that mpi4py is required by setup.py to find the right includes.  Adding mpi4py to 'requires' in pyproject.py fixes it for me - but I don't think that's the right solution.  Is there a way to specify an optional build-system dependency?

ocefpaf:
> adding --no-build-isolation to pip install seems to fix it. The test using netcdf-c github master still fails, but that's a different issue.

I wonder if that is the best solution. I'm OK doing that when mixing with conda, but there must be a better way for Python packaging.

PS: I'll remove the workaround I added if this passes, just want to make one change at a time.

setup.py
```diff
@@ -7,6 +7,15 @@
 from setuptools.dist import Distribution
 from typing import List
 
+
+
+has_mpi4py = False
```
ocefpaf, line 12: I don't like this logic. Ideally we should invert this. If netcdf-c has parallel support, we recommend users to install `mpi4py` and we do so in the tests. What do you think?

jswhit: I agree there's no need to check for mpi4py if netcdf-c doesn't have parallel support - so it is a bit backwards.

ocefpaf: Done.

ocefpaf:
This one should be ready. More changes than I would like, but the main ones are:

1. removal of Python 3.8 support
2. added Python 3.13 as the "latest" the wheel test matrix
3. added `mpi4py` as an optional dependency

PR #1415, opened 2025-08-01
Skip cp314t for now and build cp314 on PRs

ocefpaf:
@jswhit I also had to update many of the dependencies in https://github.com/ocefpaf/netcdf-manylinux and the manylinux version to get this working with Python 3.14. The build is working now, but 2  tests are when excuted with the built wheel. Do you mind taking a look? I'm not sure I know where to start fixing them. Maybe downgrading hdf5?

Closes #1413

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -60,9 +60,9 @@ jobs:
           - os: macos-14
             arch: arm64
             CIBW_ENVIRONMENT: MACOSX_DEPLOYMENT_TARGET=14.0
-          - os: macos-12
+          - os: macos-13
             arch: x86_64
-            CIBW_ENVIRONMENT: MACOSX_DEPLOYMENT_TARGET=12.0
+            CIBW_ENVIRONMENT: MACOSX_DEPLOYMENT_TARGET=13.0
 
     steps:
     - uses: actions/checkout@v4


jswhit:
Not an HDF5 issue - it seems that numpy 2.3.2 is pickier about how it handles iterable slice indices. I will have a look.

jswhit:
hmmm - not able to reproduce this on my mac with python 3.13.5 and numpy 2.3.2

jswhit:
If you update your fork from master, the tests should now pass on python 3.14

jswhit:
Looks like macos-12 is no longer supported, macos-13 is the only one that supports x86_64

ocefpaf:
> Looks like macos-12 is no longer supported, macos-13 is the only one that supports x86_64

Thanks! I updated that and it should be OK for a review now.

PS: To make our lives easier in the future, note that these Linux builds are using the Docker images from https://github.com/ocefpaf/netcdf-manylinux/tree/35e29a405d74de914bfb9427bf4883eb9d4ced26. I did update to use a stable release of netcdf-c in the next build. Hopefully there won't be any significant changes.

Also, we should really move that repo away from my personal account to make it easier for others to find and contribute.

ocefpaf:
@jswhit do you think we should try  `cp314t`? The free-threaded Python variant?

jswhit:
I don't think the netcdf-c library is thread safe (there was some initial work done to fix this [here](https://github.com/Unidata/netcdf-c/issues/1373), but it looks unfinished).  Given this, maybe there is not much benefit?

ocefpaf:
> Given this, maybe there is not much benefit?

It is likely to fail too. I forgot about that.

PR #1427, opened 2025-09-15
WHL: build limited-api compliant wheels

neutrinoceros:
This contains a couple separable changes:

- enable building limited-API compliant wheels
- make CI actually produce abi3 wheels (intended for both testing and publication)
- move as much configuration as possible for `cibuildwheel` into `pyproject.toml`

Note that abiding to the limited API drastically improves wheels portability, but *may* incur a cost in performance. In my experience so far with packages using Cython+numpy to build extensions, I haven't seen any measurable performance regression anywhere, but I still strongly advise experts to run benchmarks before blessing the PR as the *default* way to build wheels for this package.


ocefpaf:
Python 3.9 EoL is next month. What do you all think if we drop it now in this PR?

pyproject.toml
```diff
@@ -102,3 +103,56 @@ module = [
...
+inherit.environment = "append"
+environment = {HDF5_DIR='C:\\\\Users\\runneradmin\\micromamba\\envs\\build\\Library', netCDF4_DIR='C:\\\\Users\\runneradmin\\micromamba\\envs\\build\\Library', PATH='C:\\\\Users\\runneradmin\\micromamba\\envs\\build\\Library\\bin;${PATH}' }
+
+[[tool.cibuilwheel.overrides]]
+select = "*-win_*"
+inherit.skip = "append"
+skip = [
+  # cannot build cftime for this target (missing a wheel at the time of writing)
+  "cp314",
+]
```
neutrinoceros, line 158: this doesn't seem to work either. Could this be a bug that cibuildwheel ignores it ?

neutrinoceros:
> Python 3.9 EoL is next month. What do you all think if we drop it now in this PR?

As far as I'm concerned I'd rather do it in a separate PR, but I'm happy to contribute this too, either now or after this PR is ready (and merged ?).

neutrinoceros:
I'm also working on cftime to modernize infrastructure before I attempt to do the same thing there: https://github.com/Unidata/cftime/pull/371

pyproject.toml
neutrinoceros: reported as https://github.com/pypa/cibuildwheel/issues/2596

neutrinoceros:
The remaining failures seem unrelated (stubtests + a network error), so I'll undraft now

ocefpaf:
@jswhit this is highly desirable IMO, it will reduce the build matrix and make it easier to maintain them. We could do some dev releases if you want to be careful and have more people testing the performance impacts. However, I'm 99% sure they'll be minor and the benefits will outweigh the disadvantages.

PS: I would remove py39 here b/c we are unlikely to mint a new version release before its EoL. It will reduce the matrix even further and remove on of the matrix items where ABI3 isn't possible.

ocefpaf (approved):

neutrinoceros:
Rebased to resolve merge conflicts, and added a commit to drop 3.9

neutrinoceros:
Should I also add a `'Programming Language :: Python :: 3.14'` PyPI classifier while I'm at it ?

jswhit:
good idea, please do

neutrinoceros:
done !

PR #1436, opened 2025-10-13
attempt to fix wheel building logic 

jswhit:
only wheels for 3.10 and 3.11 were being built


jswhit:
@neutrinoceros I copied this from cftime - does this look right to you?

neutrinoceros:
Thanks for the ping; it doesn't. Building only `cp310-cp310` and `cp311-abi3` wheels was the intention:
- the former are still needed for Python 3.10 because Cython isn't able to generate abi3 extensions when using numpy (it really wants to use Py_Buffer, new in 3.11)
- the latter is forward compatible, so no builds are needed beyond it

jswhit:
oh wow - I had no idea.  I'll leave it as is then.

PR #1437, opened 2025-10-13
use native linux arm64 runner for aarch64 wheels

jswhit:


jswhit:
doesn't work since manylinux image not available

PR #1448, opened 2025-11-10
fix inverted conditions for plugin detection

jswhit:


jswhit:
windows tests for compression plugins are failing - don't know why

PR #1484, opened 2026-07-07
Bump osx-arm64 wheel to latest macos-15

ocefpaf:
Testing if this builds without any issues.

### DIFF
--- a/.github/workflows/cibuildwheel.yml
+++ b/.github/workflows/cibuildwheel.yml
@@ -104,7 +104,7 @@ jobs:
         include:
           - os: windows-latest
             arch: AMD64
-          - os: macos-14
+          - os: macos-15
             arch: arm64
           - os: macos-15-intel
             arch: x86_64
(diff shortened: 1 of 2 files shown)


ocefpaf:
Looks like it works for arm64 on macos-15.

@jswhit do you mind taking a look and, if possible, mint a release so we can get this new wheel out there?

ocefpaf:
Maybe hold this one. I was mixing conda python built with modern macos, but for wheels we are probably OK with the oldest one available.

ocefpaf:
@jswhit I re-run this PR just to test the latest manylinux docker image and things worked fine. In the that image we have:

```python
In [1]: import netCDF4
In [2]: netCDF4.__hdf5libversion__
Out[2]: '2.2.0'

In [3]: netCDF4.__netcdf4libversion__
Out[3]: '4.10.1'
```

jswhit:
is this one ready to merge?

ocefpaf:
> is this one ready to merge?

I would wait a bit. I guess that we can do the next wheel with the older image first.

